PR #1, opened 2022-03-17
Added intsuffix

ZuluPro:
Added intsuffix to humanize into just the symbol k, M, G, P, T, etc

### DIFF
--- a/src/humanize/__init__.py
+++ b/src/humanize/__init__.py
@@ -8,6 +8,7 @@ from humanize.number import (
     fractional,
     intcomma,
     intword,
+    intsuffix,
     ordinal,
     scientific,
 )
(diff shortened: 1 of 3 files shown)


hugovk:
Thanks for the PR! Please also add it to `__all__` in https://github.com/python-humanize/humanize/blob/master/src/humanize/__init__.py, that should fix the tests.

hugovk:
Thinking about the name `intsuffix`: what we're doing here is taking a large integer, and finding a good ["unit prefix"](https://en.wikipedia.org/wiki/Unit_prefix) to go with it.

Would `intprefix` be a better name?

hugovk:
Closing in favour of https://github.com/python-humanize/humanize/pull/22 which should more-or-less cover this too. Thanks!

PR #15, opened 2022-05-03
Add type hints

hugovk:

* Redo of https://github.com/jmoiron/humanize/pull/150
* Should be a major release because some of the returns have changed from `return value` to `return str(value)`

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -4,9 +4,11 @@
 
 These are largely borrowed from Django's `contrib.humanize`.
 """
+from __future__ import annotations
 
 import datetime as dt
 import math
+import typing
 from enum import Enum
 from functools import total_ordering
 
@@ -34,17 +36,17 @@ class Unit(Enum):
     MONTHS = 6
     YEARS = 7
 
-    def __lt__(self, other):
+    def __lt__(self, other: typing.Any) -> typing.Any:
         if self.__class__ is other.__class__:
             return self.value < other.value
         return NotImplemented
 
 
-def _now():
+def _now() -> dt.datetime:
     return dt.datetime.now()
 
 
-def _abs_timedelta(delta):
+def _abs_timedelta(delta: dt.timedelta) -> dt.timedelta:
     """Return an "absolute" value for a timedelta, always representing a time distance.
 
     Args:
@@ -59,7 +61,9 @@ def _abs_timedelta(delta):
     return delta
 
 
-def _date_and_delta(value, *, now=None):
+def _date_and_delta(
+    value: typing.Any, *, now: dt.datetime | None = None
+) -> tuple[typing.Any, typing.Any]:
     """Turn a value into a date and a timedelta which represents how long ago it was.
 
     If that's not possible, return `(None, value)`.
@@ -83,7 +87,7 @@ def _date_and_delta(value, *, now=None):
 
 
 def naturaldelta(
-    value,
+    value: dt.timedelta | int,
     months: bool = True,
     minimum_unit: str = "seconds",
 ) -> str:
@@ -122,7 +126,7 @@ def naturaldelta(
     tmp = Unit[minimum_unit.upper()]
     if tmp not in (Unit.SECONDS, Unit.MILLISECONDS, Unit.MICROSECONDS):
         raise ValueError(f"Minimum unit '{minimum_unit}' not supported")
-    minimum_unit = tmp
+    min_unit = tmp
 
     if isinstance(value, dt.timedelta):
         delta = value
@@ -131,7 +135,7 @@ def naturaldelta(
             value = int(value)
             delta = dt.timedelta(seconds=value)
         except (ValueError, TypeError):
-            return value
+            return str(value)
 
     use_months = months
 
@@ -139,22 +143,21 @@ def naturaldelta(
     days = abs(delta.days)
     years = days // 365
     days = days % 365
-    months = int(days // 30.5)
+    num_months = int(days // 30.5)
 
     if not years and days < 1:
         if seconds == 0:
-            if minimum_unit == Unit.MICROSECONDS and delta.microseconds < 1000:
+            if min_unit == Unit.MICROSECONDS and delta.microseconds < 1000:
                 return (
                     _ngettext("%d microsecond", "%d microseconds", delta.microseconds)
                     % delta.microseconds
                 )
-            elif minimum_unit == Unit.MILLISECONDS or (
-                minimum_unit == Unit.MICROSECONDS
-                and 1000 <= delta.microseconds < 1_000_000
+            elif min_unit == Unit.MILLISECONDS or (
+                min_unit == Unit.MICROSECONDS and 1000 <= delta.microseconds < 1_000_000
             ):
                 milliseconds = delta.microseconds / 1000
                 return (
-                    _ngettext("%d millisecond", "%d milliseconds", milliseconds)
+                    _ngettext("%d millisecond", "%d milliseconds", int(milliseconds))
                     % milliseconds
                 )
             return _("a moment")
@@ -178,23 +181,24 @@ def naturaldelta(
         if not use_months:
             return _ngettext("%d day", "%d days", days) % days
         else:
-            if not months:
+            if not num_months:
                 return _ngettext("%d day", "%d days", days) % days
-            elif months == 1:
+            elif num_months == 1:
                 return _("a month")
             else:
-                return _ngettext("%d month", "%d months", months) % months
+                return _ngettext("%d month", "%d months", num_months) % num_months
     elif years == 1:
-        if not months and not days:
+        if not num_months and not days:
             return _("a year")
-        elif not months:
+        elif not num_months:
             return _ngettext("1 year, %d day", "1 year, %d days", days) % days
         elif use_months:
-            if months == 1:
+            if num_months == 1:
                 return _("1 year, 1 month")
             else:
                 return (
-                    _ngettext("1 year, %d month", "1 year, %d months", months) % months
+                    _ngettext("1 year, %d month", "1 year, %d months", num_months)
+                    % num_months
                 )
         else:
             return _ngettext("1 year, %d day", "1 year, %d days", days) % days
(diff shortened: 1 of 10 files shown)


pawamoy:
Hmmm, I was going to say ["get_type_hints + pytkdocs won't work on python less than 3.9"](https://github.com/mkdocstrings/pytkdocs/issues/139#issuecomment-1083430183) but your pipeline ran on python 3.10 :thinking:

hugovk:
`typing.TypeAlias` is new in 3.10, maybe mkdocstrings / pytkdocs doesn't support it yet?

https://docs.python.org/3/library/typing.html#typing.TypeAlias

pawamoy:
Maybe it's simply because your `TypeAlias` import is type guarded?

hugovk:
> Maybe it's simply because your `TypeAlias` import is type guarded? Therefore not available at runtime for `get_type_hints` to pick it up? pytkdocs doesn't do anything special with types (it has no particular knowledge of any other types).

Removing `if TYPE_CHECKING:` fixes it. I'd rather not remove it, because it would mean we need to add `typing_extensions` as a runtime dependency, when it's only needed when checking types.

> Could you try to patch pytkdocs' loader module to add the future annotations import and see if it keeps failing?

This doesn't fix it.

tests/test_time.py
```diff
@@ -599,27 +613,27 @@ def test_precisedelta_custom_format(val, min_unit, fmt, expected):
...
-def test_time_unit():
+def test_time_unit() -> None:
     years, minutes = time.Unit["YEARS"], time.Unit["MINUTES"]
     assert minutes < years
     assert years > minutes
     assert minutes == minutes
 
     with pytest.raises(TypeError):
-        years < "foo"
+        assert years < "foo"
```
coiax, line 639: Assume this change is to avoid some sort of linter complaining about an expression that does nothing.

But the assertion implies you want to check the truthish of the expression.

What about 
```suggestion
        _ = years < "foo"
```

Using `_` as the generic "We don't care about this" variable. Is the linter happy with that?

tests/test_number.py
```diff
@@ -170,5 +175,5 @@ def test_scientific(test_args, expected):
         ([1, humanize.intword, 1e6, None, "under "], "under 1.0 million"),
     ],
 )
-def test_clamp(test_args, expected):
+def test_clamp(test_args: list[typing.Any], expected: str) -> None:
```
coiax, line 178: For nicer type hints, you can put

```
if typing.TYPE_CHECKING:
    from typing import Any
```
at the top of the file, and just go `list[Any]` here.

But that's a stylistic choice, nothing wrong with this either.

tests/test_filesize.py
```diff
@@ -32,7 +33,7 @@
         ([10**26 * 30, True, False, "%.3f"], "2481.542 YiB"),
     ],
 )
-def test_naturalsize(test_args, expected):
+def test_naturalsize(test_args: list[int] | list[int | bool], expected: str) -> None:
```
coiax, line 36: This is a very specific type hint for something that's just going to be slapped into argument unpacking. 😅

src/humanize/time.py
```diff
@@ -396,7 +413,12 @@ def _suppress_lower_units(min_unit, suppress):
     return suppress
 
 
-def precisedelta(value, minimum_unit="seconds", suppress=(), format="%0.2f") -> str:
+def precisedelta(
+    value: dt.timedelta | int,
+    minimum_unit: str = "seconds",
+    suppress: typing.Iterable[str] = (),
```
coiax, line 419: Use `collections.abc.Iterable` instead. Using it from `typing` has been [deprecated](https://docs.python.org/3/library/typing.html#typing.Iterable) since 3.9, although we've got 5 years to make the change.

src/humanize/time.py
```diff
@@ -312,14 +322,21 @@ def _quotient_and_remainder(value, divisor, unit, minimum_unit, suppress):
...
+        return 0, value
     else:
         return divmod(value, divisor)
 
 
-def _carry(value1, value2, ratio, unit, min_unit, suppress):
+def _carry(
+    value1: int | float,
+    value2: int | float,
+    ratio: int | float,
```
coiax, line 335: It's okay to just have `float` here, Python is fine with passing an `int` to a float argument, and we're likely doing `float` operations on them anyway.

coiax (commented):
Looks good, couple of observations, but none of them blocking. :+1:

tests/test_time.py
hugovk: Yeah, just a silent warning in PyCharm:

image


image

It's happy with your suggestion:

image

tests/test_filesize.py
hugovk: Yeah, it's a bit of a jumble! Will replace with `list[typing.Any]`

src/humanize/time.py
hugovk: Good to know, will replace.

src/humanize/time.py
hugovk: Thanks, will replace all the `int | float` with `float.

tests/test_filesize.py
hugovk: Hmm, with this:
```diff
diff --git a/tests/test_filesize.py b/tests/test_filesize.py
index 0119d58..857a958 100644
--- a/tests/test_filesize.py
+++ b/tests/test_filesize.py
@@ -3,6 +3,8 @@
 """Tests for filesize humanizing."""
 from __future__ import annotations
 
+import typing
+
 import pytest
 
 import humanize
@@ -33,7 +35,7 @@ import humanize
         ([10**26 * 30, True, False, "%.3f"], "2481.542 YiB"),
     ],
 )
-def test_naturalsize(test_args: list[int] | list[int | bool], expected: str) -> None:
+def test_naturalsize(test_args: list[typing.Any], expected: str) -> None:
     assert humanize.naturalsize(*test_args) == expected
 
     args_with_negative = test_args
```

The mypy fails when run from pre-commit but _only as part of the Git commit_:

```console
$ gc -m "Use typing.Any"
pyupgrade................................................................Passed
black....................................................................Passed
isort....................................................................Passed
autoflake................................................................Passed
flake8...................................................................Passed
check blanket noqa.......................................................Passed
check for merge conflicts................................................Passed
check toml...........................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
fix end of files.........................................................Passed
pydocstyle...........................................(no files to check)Skipped
- hook id: mypy
- exit code: 1

tests/test_filesize.py:10: error: Cannot find implementation or library stub for module named "humanize"
tests/test_filesize.py:10: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)

setup-cfg-fmt........................................(no files to check)Skipped
pyproject-fmt........................................(no files to check)Skipped
tox-ini-fmt..........................................(no files to check)Skipped
```

Fine when run in pre-commit or directly as mypy:

```console
$ pre-commit run --all-files
pyupgrade................................................................Passed
black....................................................................Passed
isort....................................................................Passed
autoflake................................................................Passed
flake8...................................................................Passed
check blanket noqa.......................................................Passed
check for merge conflicts................................................Passed
check toml...............................................................Passed
check yaml...............................................................Passed
fix end of files.........................................................Passed
pydocstyle...............................................................Passed
mypy.....................................................................Passed
setup-cfg-fmt............................................................Passed
pyproject-fmt............................................................Passed
tox-ini-fmt..............................................................Passed
$ mypy --strict .
Success: no issues found in 11 source files
```

Maybe I'll just leave the jumble...

src/humanize/time.py
pawamoy: Or use `typing.Sequence`? Not sure this would fit here, just my 2 cents 🙂

pawamoy:
> Removing if TYPE_CHECKING: fixes it. I'd rather not remove it, because it would mean we need to add typing_extensions as a runtime library dependency, when it's only needed when checking types.

Of course, completely understandable.

> This doesn't fix it.

Erm, too bad. I'm not sure what is required at this point to make `get_type_hints` work. I'll need to investigate a bit more. Thank you for trying!

hugovk:
@pawamoy Hi! Any news on this? Do you have an upstream issue to follow?

pawamoy:
No news, no issue. Didn't take the time to investigate. At this point I'd recommend trying the new handler, which does not use introspection but rather visits the AST (I don't know if you already tried it or not).

hugovk:
I didn't try the new handler but happy to give it a go! What do we need to change to try it?

pawamoy:
You just need to depend on `mkdocstrings-python` instead of `mkdocstrings-python-legacy`, or to use the `python` extra instead of the `python-legacy` one :slightly_smiling_face: 

More information about the two handlers: https://mkdocstrings.github.io/handlers/overview/#about-the-python-handlers.
Highlight: the new handler does not support documenting inherited members yet, and there are known issues with wildcard imports.

hugovk:
Thanks, that works!

hugovk:
@sodul In which function? And what version of Humanize are you using?

sodul:
@hugovk We got the issue with `humanize==4.2.3`, rolling back to `humanize==4.1.0` helped.

Sample error:
```
error: Argument 1 to "precisedelta" has incompatible type "float"; expected "Union[timedelta, int]"
```

I took an other look and in a way the `int` type properly reflects the actual behavior:
```
>>> import humanize
>>> humanize.precisedelta(0.05)
'0 seconds'

>>> humanize.precisedelta(1.05)
'1 second'

>>> humanize.precisedelta(0.05, minimum_unit="microseconds")
'0 microseconds'

>>> humanize.precisedelta(1.05, minimum_unit="microseconds")
'1 second'

>>> humanize.precisedelta(1.05, minimum_unit="milliseconds")
'1 second'

>>> humanize.precisedelta(0.05, minimum_unit="milliseconds")
'0 milliseconds'
```

When the seconds are passed as a float it works, but only the integer part is used. That's because `_date_and_delta()` does `value = int(value)` but I think that it could do `value = float(value)` instead, and then only if the type is not already a number (I suspect the intent was to cast strings to a number value).

So at the end of the day ... the type annotation is reflecting the behavior. Personally I believe `float` is fine (it accepts `int` types as a subset of `float`), and that the code could even benefit to actually support fractions of seconds when `minimum_unit` allows for it.

nuztalgia:
@sodul I have an open PR (#39) that I believe will fix the mypy issue. It also adds more precise support for float values:
```py
>>> precisedelta(0.01, "seconds", format="%0.3f")
BEFORE:  "0 seconds"
AFTER:   "0.010 seconds"
```

PR #16, opened 2022-05-04
Remove redundant `wheel` dependency from `pyproject.toml`

mgorny:
The `wheel` dependency in `pyproject.toml` is not necessary, and modern
setuptools documentation advises against adding it.  The PEP517 backend
automatically exposes the `wheel` dependency, and a future version
of setuptools may no longer use it.


hugovk:
Thanks!

A couple of references:

* https://github.com/pypa/setuptools_scm/pull/689
* https://github.com/pypa/setuptools/pull/3056

PR #19, opened 2022-05-16
Fix days to hours boundary for precise delta

### LINKED ISSUE #14: Issues with odd months when restricting to "day" precision (precisedelta)
### What did you do?

I'm trying to provide more precision than "months" when the `timedelta` is more than 1 month, but I'm finding some issues using `precisedelta` set to `days` with a format of `%d` as seen below.  This all seems to stem from the use of 30.5 as a divisor for mod in `precisedelta` and how it and `ngettext` interact with floats vs. ints when it comes to pluralize and the display of the days itself.  For what it's worth, the same thing happens at second precision with microseconds present, etc.  It seems to be at the `minimum_unit` where this logic takes place.

```python
>>> import humanize
>>> import datetime
>>> humanize.precisedelta(datetime.timedelta(days=31), minimum_unit="days", format="%d")
'1 month'
>>> 
```
```python
>>> import humanize
>>> import datetime
>>> humanize.precisedelta(datetime.timedelta(days=92), minimum_unit="days", format="%d")
'3 months'
>>> 
```
```python
>>> import humanize
>>> import datetime
>>> humanize.precisedelta(datetime.timedelta(days=32), minimum_unit="days", format="%d")
'1 month and 1 day'
>>>
```

### What actually happened?

```python
>>> import humanize
>>> import datetime
>>> humanize.precisedelta(datetime.timedelta(days=31), minimum_unit="days", format="%d")
'1 month and 0 days'
>>> 
```
```python
>>> import humanize
>>> import datetime
>>> humanize.precisedelta(datetime.timedelta(days=92), minimum_unit="days", format="%d")
'3 month and 0 days'
>>> 
```
```python
>>> import humanize
>>> import datetime
>>> humanize.precisedelta(datetime.timedelta(days=32), minimum_unit="days", format="%d")
'1 month and 1 days'
>>>
```

* OS: Windows, Ubuntu 20.04
* Python: 3.8.10
* Humanize: 4.1.0

Please include **code** that reproduces the issue (included in expected, actually happened above)

eldipa:
Fixes #14 

* Propagate correctly the days to hours in `precisedelta`
* Propose a `truncate` flag to truncate and remove values too close to zero in `precisedelta`

Note: I don't feel that `truncate` is the correct name and I think that it could be easily misinterpreted. We need a better name.
This PR also does not include a documentation for this flag nor tests (but a file in `tests/issue-14-days-hours-boundary-in-precisedelta.md` describes and tests all the cases discussed in #14 so it can be used as the base for an unit test and/or as the tests themselves running them with [byexample](https://byexamples.github.io/byexample/) )

Closes #14

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -39,6 +39,20 @@ class Unit(Enum):
             return self.value < other.value
         return NotImplemented
 
+    def succ(self):
+        if self == Unit.YEARS:
+            raise IndexError()
+
+        v = self.value + 1
+        return Unit(v)
+
+    def pred(self):
+        if self == Unit.MICROSECONDS:
+            raise IndexError()
+
+        v = self.value - 1
+        return Unit(v)
+
 
 def _now():
     return dt.datetime.now()
@@ -396,7 +410,7 @@ def _suppress_lower_units(min_unit, suppress):
     return suppress
 
 
-def precisedelta(value, minimum_unit="seconds", suppress=(), format="%0.2f") -> str:
+def precisedelta(value, minimum_unit="seconds", suppress=(), format="%0.2f", truncate=False) -> str:
     """Return a precise representation of a timedelta.
 
     ```pycon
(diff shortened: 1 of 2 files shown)


dangillet:
@hugovk I think this PR could be closed. https://github.com/python-humanize/humanize/pull/254 was merged and superseded this one.

PR #22, opened 2022-06-17
Add humanize.metric() for converting big/small numbers to SI units

bwoodsend:
Fixes nothing!

* Fix scientific() on small positive numbers.
* Add humanize.metric() for converting big/small numbers to SI units.

### DIFF
--- a/src/humanize/__init__.py
+++ b/src/humanize/__init__.py
@@ -8,6 +8,7 @@ from humanize.number import (
     fractional,
     intcomma,
     intword,
+    metric,
     ordinal,
     scientific,
 )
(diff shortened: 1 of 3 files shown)


bwoodsend:
Uhm, why did precommit do that?

hugovk:
Haha! That's autoflake fixing:

```
src/humanize/__init__.py:5:1: F401 'humanize.number.metric' imported but unused
```

The real fix is to add `metric` to `__all__` in `src/humanize/__init__.py`.

hugovk:
Thanks for this!

For `metric` let's add the required space between the number and the unit (maybe with exceptions for plane angular degrees, minutes, and seconds (°, ′, and ″)?).

https://en.wikipedia.org/wiki/International_System_of_Units#General_rules

I guess we should do this also when there's no SI unit but we end up with a metric `k` or so on?

I think this will replace PR https://github.com/python-humanize/humanize/pull/1 as well.

hugovk:
Thank you!

PR #23, opened 2022-06-20
BUG: Use %d for year translations convert to string for intcomma after

### LINKED ISSUE #21: naturaltime:  wrong translation with delta >= 2 years
### What did you do?

Ran the following code:

```python
import humanize
import datetime as dt

output = humanize.naturaltime(dt.datetime(year=2010, month=1, day=1))
print(output)
```

Output to be:
```
il y a 12 ans
```
### What actually happened?

Output was:
```
il y a 12 years
```

* OS: Mac OS
* Python: 3.8.6
* Humanize: 4.0.1

carterbox:
This patch fixes a bug introduced in 3.14.0, where the format
string was changed from %d to %s to add separators to the year.
However, this needs to happen after translation because the
translator uses the format strings as part of the translation.

Closes #21

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -202,7 +202,7 @@ def naturaldelta(
         else:
             return _ngettext("1 year, %d day", "1 year, %d days", days) % days
 
-    return _ngettext("%s year", "%s years", years) % intcomma(years)
+    return _ngettext("%d year", "%d years", years).replace("%d", "%s") % intcomma(years)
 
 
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks for the fix!

Please could you also add a test to `test_i18n.py`? For either `naturaltime` or `naturaldelta`.

hugovk:
Done! Will set to draft as well.

hugovk:
Thank you!

PR #41, opened 2022-07-08
Fix intword for negative numbers

### LINKED ISSUE #12: intword with negative numbers
### What did you do?

```python
import humanize
print(humanize.intword(-1000000000))
```

```
-1.0 billion
```

### What actually happened?

```
-1000000000
```

* OS: Windows 10
* Python: 3.10.4
* Humanize: 3.14

vishket:
Fixes #12 

* Support `intword` for negative numbers
* Update corresponding test `test_intword`


src/humanize/number.py
```diff
@@ -205,23 +205,44 @@ def intword(value: NumberOrString, format: str = "%.1f") -> str:
...
+                    return (
+                        "-"
+                        + " ".join(
+                            [format, _ngettext(singular, plural, math.ceil(chopped))]
+                        )
+                    ) % chopped
+
                 return (
                     " ".join([format, _ngettext(singular, plural, math.ceil(chopped))])
                 ) % chopped
```
carterbox, line 231:
```suggestion
                return (
                    "-" if is_negative else ""
                    " ".join([format, _ngettext(singular, plural, math.ceil(chopped))])
                ) % chopped
```
I think something like this would work because strings within parenthesis are concatenated? It would reduce duplicated code. You could do something similar at every place you need to add a negative sign to the output.

carterbox (changes requested):

hugovk (changes requested):
A suggestion :)

hugovk:
Hmm, looks like my suggestion didn't post, will try again!

src/humanize/number.py
```diff
@@ -205,23 +205,44 @@ def intword(value: NumberOrString, format: str = "%.1f") -> str:
     except (TypeError, ValueError):
         return str(value)
 
+    is_negative = value < 0
+
+    if is_negative:
+        value *= -1
```
hugovk, line 211: How about something like this?

```suggestion
    if value < 0:
        value *= -1
        negative_prefix = "-"
    else:
        negative_prefix = ""
```

Then we can remove the `if is_negative` bits (and some duplication) and always prepend `negative_prefix` to the return?

src/humanize/number.py
```diff
@@ -205,23 +205,44 @@ def intword(value: NumberOrString, format: str = "%.1f") -> str:
...
+
     if value < powers[0]:
-        return str(value)
+        return "-" + str(value) if is_negative else str(value)
```
hugovk, line 214: For example:

```suggestion
        return negative_prefix + str(value)
```

And similarly below.

hugovk (changes requested):
Take 2!

src/humanize/number.py
vishket: @hugovk much cleaner! thanks for the suggestion

hugovk (approved):

hugovk:
Thank you @vishket for the PR and @carterbox for the review!

PR #47, opened 2022-08-08
Fix "ValueError: math domain error" for metric(0)

liukun:
Call metrics(0) gives `ValueError: math domain error` before this patch.


hugovk:
Thanks!

PR #52, opened 2022-08-25
Fix intcomma() failing with a string as input when ndigits is not None

Luflosi:
Calling `humanize.intcomma("1", 0)` would fail with the error message
```
ValueError: Unknown format code 'f' for object of type 'str'
```

Fix this by first converting the string to a `float` or `int`. Always converting the string to a `float` would not work for cases like `humanize.intcomma("1")`, as this would output `'1.0'` instead of the desired output `'1'`.
This also requires using a while loop instead of recursion, as calling this function again would now remove the thousands separator, leading to an infinite recursion. This also makes it easier to reason about IMO.

* Fix error described above
* Add some new test cases
* Add a new example


hugovk:
Thank you!

Issue #48, opened 2022-08-13
German localisation should be improved


Luflosi:
### What did you do?
Execute
```python
import humanize
import datetime as dt
_t = humanize.i18n.activate("de_DE")
humanize.naturaltime(dt.datetime.now() - dt.timedelta(seconds=3600))
```

### What did you expect to happen?
I would expect the output to be 'vor einer Stunde'.

### What actually happened?
The output was 'vor eine Stunde', which is grammatically incorrect.
"one hour" translates to "eine Stunde" but "one hour ago" translates to "vor einer Stunde". I think 1 is the only number where this special case happens.

### What versions are you using?

* OS: macOS 10.15.7
* Python: 3.10.5
* Humanize: 4.2.3

Overall the support for German seems a bit lacking.
For example the `.` and `,` in numbers are swapped compared to what's common in English, which this library doesn't support (`humanize.intcomma(12345)` for example).
`humanize.intword(123455913)` outputs `123.5 Million` but `Million` is singular. The correct output with a plural would be `123.5 Millionen`, since 123.5 is more than 1 (actually I think it's correct to always use the plural when the value is not exactly equal to 1).

Is this something you think would be worthwhile to improve? I could help you by trying to find more of these problems, testing patches or even writing some fixes myself if you help me a bit. I'm afraid if I just mess around in the translation file, I'll fix some specific problem I found but break other cases I didn't think about.

carterbox:
The translation files are located here: https://github.com/python-humanize/humanize/tree/main/src/humanize/locale

I'm not sure whether they can handle a special case for single number like that because the pattern matches for "%s ago". Maybe poke around the language files and see if there are other languages that have a solution for this.

> humanize.intword(123455913) outputs 123.5 Million but Million is singular. The correct output with a plural would be 123.5 Millionen, since 123.5 is more than 1 (actually I think it's correct to always use the plural when the value is not exactly equal to 1).

The translation files do support this type of plural selection. It probably just needs to be implemented in the translation file for German ([see Spanish for an example](https://github.com/python-humanize/humanize/blob/14fc392cf49936fee06edd41863b1372ffc29d95/src/humanize/locale/ca_ES/LC_MESSAGES/humanize.po#L158)).

> For example the . and , in numbers are swapped compared to what's common in English, which this library doesn't support (humanize.intcomma(12345) for example).

[...]

PR #53, opened 2022-08-30
Internationalise the decimal separator in intcomma()

Luflosi:
See discussion in #48.

* Add the proper decimal separator for the de_DE locale
* Add some tests

### DIFF
--- a/src/humanize/__init__.py
+++ b/src/humanize/__init__.py
@@ -1,7 +1,7 @@
 """Main package for humanize."""
 
 from humanize.filesize import naturalsize
-from humanize.i18n import activate, deactivate, thousands_separator
+from humanize.i18n import activate, deactivate, decimal_separator, thousands_separator
 from humanize.number import (
     apnumber,
     clamp,
@@ -36,6 +36,7 @@ __all__ = [
     "apnumber",
     "clamp",
     "deactivate",
+    "decimal_separator",
     "fractional",
     "intcomma",
     "intword",
--- a/src/humanize/i18n.py
+++ b/src/humanize/i18n.py
@@ -5,7 +5,7 @@ import gettext as gettext_module
 import os.path
 from threading import local
 
-__all__ = ["activate", "deactivate", "thousands_separator"]
+__all__ = ["activate", "deactivate", "decimal_separator", "thousands_separator"]
 
 _TRANSLATIONS: dict[str | None, gettext_module.NullTranslations] = {
     None: gettext_module.NullTranslations()
@@ -17,6 +17,11 @@ _CURRENT = local()
 _THOUSANDS_SEPARATOR = {
     "de_DE": ".",
     "fr_FR": " ",
+}
+
+# Mapping of locale to decimal separator
+_DECIMAL_SEPARATOR = {
+    "de_DE": ",",
 }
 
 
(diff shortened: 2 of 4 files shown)


tests/test_i18n.py
```diff
@@ -44,6 +44,12 @@ def test_intcomma() -> None:
...
         humanize.i18n.activate("de_DE")
         assert humanize.intcomma(number) == "10.000.000"
+        assert humanize.intcomma(1_234_567.8901) == "1.234.567,8901"
+        assert humanize.intcomma(1_234_567.89) == "1.234.567,89"
+        assert humanize.intcomma("1234567,89") == "1.234.567,89"
+        assert humanize.intcomma("1.234.567,89") == "1.234.567,89"
+        assert humanize.intcomma("1.234.567,8") == "1.234.567,8"
+
         humanize.i18n.activate("fr_FR")
         assert humanize.intcomma(number) == "10 000 000"
```
carterbox, line 54:
```
humanize.i18n.activate("EN")
assert humanize.intcomma(1_234_567.8901) == "1,234,567.8901"
```
It would be good to add a test for a non-German locale to show that the original behavior is preserved. @hugovk, are the doctests verified by the CI?

Also, it should be documented that if the input is a string, it must already match the locale. i.e. This function is not to be used to convert a string from one locale to another.

Luflosi: > It would be good to add a test for a non-German locale

There are already [tests for the default locale](https://github.com/python-humanize/humanize/blob/d8e27393dbf4ed3645ffc3464c9c44f4d8e47534/tests/test_number.py#L34-L71). Or do you mean a non-German, non-default locale?

> Also, it should be documented that if the input is a string, it must already match the locale.

This was already the case with the thousands separator btw. Where would be a good place to document this?

carterbox (approved):
Implementation closely follows existing implementation for thousands separation localization. Tests are sufficient. LGTM!

tests/test_i18n.py
Luflosi: I'm not sure how best to write this, given that it doesn't really even mention that strings are allowed, except in two examples and partially in the arguments.

hugovk: > It would be good to add a test for a non-German locale to show that the original behavior is preserved. @hugovk, are the doctests verified by the CI?

Yes, doctests are run by pytest and pytest is run on the CI.

src/humanize/__init__.py
```diff
@@ -49,4 +49,5 @@
     "precisedelta",
     "scientific",
     "thousands_separator",
+    "decimal_separator",
```
hugovk, line 52: Let's keep these in alphabetical order, makes them easier to find.

src/humanize/i18n.py
```diff
@@ -5,7 +5,7 @@
 import os.path
 from threading import local
 
-__all__ = ["activate", "deactivate", "thousands_separator"]
+__all__ = ["activate", "deactivate", "thousands_separator", "decimal_separator"]
```
hugovk, line 8: Alphabetical here too:
```suggestion
__all__ = ["activate", "deactivate", "decimal_separator", "thousands_separator"]
```

hugovk (changes requested):
Thank you for this! Just a few little nits :)

Luflosi:
Done.

hugovk (approved):

PR #68, opened 2022-10-26
Update test.yml

Ishikasaraf20:
change 3.11 dev to 3.11

Fixes #

* 
* 
*


hugovk:
Thanks for the PR. I went with https://github.com/python-humanize/humanize/pull/69 instead because it had a more useful title, description and commit message. Thanks again!

PR #76, opened 2022-11-09
Filesize simplification

bbolli:

* better test coverage filesize.py
* code simplification

### DIFF
--- a/src/humanize/filesize.py
+++ b/src/humanize/filesize.py
@@ -34,6 +34,10 @@ def naturalsize(
         '2.930K'
         >>> naturalsize(3000, True)
         '2.9 KiB'
+        >>> naturalsize(10**28)
+        '10000.0 YB'
+        >>> naturalsize(-4096, True)
+        '-4.0 KiB'
 
         ```
     Args:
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks for the PR!

Do you know why the type checking fails?

```
- hook id: mypy
- exit code: 1

src/humanize/filesize.py:80: error: Returning Any from function declared to return "str"  [no-any-return]
Found 1 error in 1 file (checked 11 source files)
```

bbolli:
> Do you know why the type checking fails?

Not really... My local mypy installation (v0.991) does not report any type violations in `src/humanize`.

hugovk:
Hmm, the same version is used by the CI:

https://github.com/python-humanize/humanize/blob/60e3b4785d78d03c5cf23d20c9fc3323a6882dfc/.pre-commit-config.yaml#L61-L66

Also with the `--strict` option.

hugovk:
Updated to make mypy happy.

The more type-safe way is:

```python
ret = format % (base * bytes_ / unit) + s
assert isinstance(ret, str)
return ret
```

But this shorter version is good enough for here. Thanks @AlexWaygood for the tips!

hugovk:
> * better test coverage filesize.py

Well, 100% to 100% is no change, and ironically by reducing the total covered lines, it's slightly decreased the total coverage percentage 🙃 

> * code simplification

but this is appreciated :)

Thank you!

PR #80, opened 2022-12-27
Cosmetic: improve exception traceback readability

hugovk:

* Assign exception messages to a temporary variable first, to make tracebacks easier to read
* See https://github.com/henryiii/flake8-errmsg

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -124,7 +124,8 @@ def naturaldelta(
     """
     tmp = Unit[minimum_unit.upper()]
     if tmp not in (Unit.SECONDS, Unit.MILLISECONDS, Unit.MICROSECONDS):
-        raise ValueError(f"Minimum unit '{minimum_unit}' not supported")
+        msg = f"Minimum unit '{minimum_unit}' not supported"
+        raise ValueError(msg)
     min_unit = tmp
 
     if isinstance(value, dt.timedelta):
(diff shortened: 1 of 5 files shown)


src/humanize/time.py
```diff
@@ -124,7 +124,8 @@ def naturaldelta(
     """
     tmp = Unit[minimum_unit.upper()]
     if tmp not in (Unit.SECONDS, Unit.MILLISECONDS, Unit.MICROSECONDS):
-        raise ValueError(f"Minimum unit '{minimum_unit}' not supported")
+        msg = f"Minimum unit '{minimum_unit}' not supported"
```
carterbox, line 127:
```suggestion
        msg = f"Minimum unit '{minimum_unit}' not supported."
```

src/humanize/time.py
```diff
@@ -409,9 +410,8 @@ def _suitable_minimum_unit(min_unit: Unit, suppress: typing.Iterable[Unit]) -> U
             if unit > min_unit and unit not in suppress:
                 return unit
 
-        raise ValueError(
-            "Minimum unit is suppressed and no suitable replacement was found"
-        )
+        msg = "Minimum unit is suppressed and no suitable replacement was found"
```
carterbox, line 413:
```suggestion
        msg = "Minimum unit is suppressed, and no suitable replacement was found."
```

carterbox (changes requested):
Might as well fix some grammar while we're here. Otherwise, LGTM. I agree that this PR improves traceback readability.

hugovk:
I think the single sentence exception messages are fine without a full stop at the end. If there are two or more sentences, then it's needed.

A quick search of CPython source code suggests the vast majority (~95%) do not end with a full stop:

```console
$ # Ending with a full stop
$ git grep "raise .*Error(.*\.['\"])" Lib | wc -l
      79
$ # Ending with no full stop
$ git grep "raise .*Error(.*[^\.]['\"])" Lib | wc -l
    1609
```

PR #99, opened 2023-02-05
Update Bengali translation

baseplate-admin:

https://github.com/python-humanize/humanize/blob/a759dd9ca8fbd6366eff09ab7a45d3104a8e548c/src/humanize/locale/bn_BD/LC_MESSAGES/humanize.po#L216

Here `পাচ` is not a valid word in bengali. The correct translation for `five` is  `পাঁচ`

---


hugovk:
Thanks!

PR #110, opened 2023-04-21
Add natural_list

### LINKED ISSUE #103: Feature: Natural list
I was wondering if we can have a feature that converts a list to a sentence; example - `["dog", "cat", "mouse"]` and convert it to `Dog, Cat and Mouse`.

I can add this feature if you are up for it. Thanks.

akshaybabloo:
* This PR adds `natural_list` to convert a list of items into a comma separated sentence with `and` at the last.
* I have also ignored `.idea/*` folder

closes #103


akshaybabloo:
I am not sure why I am getting `NameError: name 'TypeAlias' is not defined`. Any idea?

akshaybabloo:
Could you please run the CI again? Thanks.

hugovk:
Thanks for the PR! I've made some updates, and because this is a new API, added an underscore to the function.

(I'm considering adding underscores to the others too, and having the non-underscore ones be aliases. If so, no need to remove or deprecate the old ones, at least not for a long time.)

PR #123, opened 2023-06-30
Move logic to convert aware datetime to naive into function

### LINKED ISSUE #122: TypeError: can't subtract offset-naive and offset-aware datetimes
### What did you do?

I wanted to calculate naturaltime between two timezone aware datatimes.

Get a string the with naturaltime between those datetimes.

### What actually happened?

A TypeError exception was raised:

```sh
  File "/home/erik/python/projects/aa-dev/aa-memberaudit/temp/humanize_bug.py", line 8, in <module>
    result = humanize.naturaltime(same_dt, when=my_now)
  File "/home/erik/python/projects/aa-dev/venv/lib/python3.10/site-packages/humanize/time.py", line 253, in naturaltime
    date, delta = _date_and_delta(value, now=now)
  File "/home/erik/python/projects/aa-dev/venv/lib/python3.10/site-packages/humanize/time.py", line 76, in _date_and_delta
    delta = now - value
TypeError: can't subtract offset-naive and offset-aware datetimes
```

* OS: Ubuntu 22.04.2
* Python:  3.10
* Humanize: 4.7.0

The [best reproductions](https://stackoverflow.com/help/minimal-reproducible-example) are [self-contained scripts](https://ericlippert.com/2014/03/05/how-to-debug-small-programs/) with minimal dependencies.

```python
import datetime as dt

import humanize

my_now = dt.datetime.now(tz=dt.timezone.utc)
same_dt = my_now - dt.timedelta(days=3, hours=1)

result = humanize.naturaltime(same_dt, when=my_now)
```

### Additional information

This bug was introduced with 4.7, the code works fine with 4.6.

As a workaround one could remove the `when=my_now` part (a workaround to get it the function working with aware datetimes), but then it no longer works with humanize < 4.7, which would make 4.7 a breaking change. And it would also not work for when `when` is used in any other way.

ErikKalkoken:
Fixes #122 

* Add tests for naturaltime() with when parameter and timezone aware datetimes
* Refactor logic for converting timezone aware datetimes into naive datetimes into a function

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -246,11 +246,11 @@ def naturaltime(
     Returns:
         str: A natural representation of the input in a resolution that makes sense.
     """
+    value = _convert_aware_datetime(value)
     when = _convert_aware_datetime(when)
 
     now = when or _now()
 
-    value = _convert_aware_datetime(value)
 
     date, delta = _date_and_delta(value, now=now)
     if date is None:
@@ -271,7 +271,7 @@ def naturaltime(
 def _convert_aware_datetime(
     value: dt.datetime | dt.timedelta | float | None,
 ) -> Any:
-    """Convert aware datetime to naive datetime and passthrough any other type."""
+    """Convert aware datetime to naive datetime and pass through any other type."""
     if isinstance(value, dt.datetime) and value.tzinfo is not None:
         value = dt.datetime.fromtimestamp(value.timestamp())
     return value
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -251,7 +251,6 @@ def naturaltime(
 
     now = when or _now()
 
-
     date, delta = _date_and_delta(value, now=now)
     if date is None:
         return str(value)
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -12,6 +12,7 @@ import math
 import typing
 from enum import Enum
 from functools import total_ordering
+from typing import Any
 
 from .i18n import _gettext as _
 from .i18n import _ngettext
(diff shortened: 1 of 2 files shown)


src/humanize/time.py
```diff
@@ -245,10 +246,11 @@ def naturaltime(
     Returns:
         str: A natural representation of the input in a resolution that makes sense.
     """
+    when = _convert_aware_datetime(when)
```
hugovk, line 249: Let's move this up:
```suggestion
    value = _convert_aware_datetime(value)
    when = _convert_aware_datetime(when)
```

src/humanize/time.py
```diff
@@ -245,10 +246,11 @@ def naturaltime(
...
+
     now = when or _now()
 
-    if isinstance(value, dt.datetime) and value.tzinfo is not None:
-        value = dt.datetime.fromtimestamp(value.timestamp())
+    value = _convert_aware_datetime(value)
```
hugovk, line 253:
```suggestion
```

src/humanize/time.py
```diff
@@ -266,6 +268,15 @@ def naturaltime(
     return str(ago % delta)
 
 
+def _convert_aware_datetime(
+    value: dt.datetime | dt.timedelta | float | None,
+) -> Any:
+    """Convert aware datetime to naive datetime and passthrough any other type."""
```
hugovk, line 274:
```suggestion
    """Convert aware datetime to naive datetime and pass through any other type."""
```

hugovk (commented):
Thanks for the PR, a couple of minor suggestions.

thedoctorjtd (approved):
Revisions make sense; original changes should fix the issue.

PR #130, opened 2023-07-09
* time.py, added iso 8601 support

velle:
* time.py, added new function _parseiso(value)* relevant functions now call value = _parseiso(value)
* wrote inline documentation, and short subparagraph to README.md
* have not tested code in any way except the existing pytests which all succeeded
* tox gives mypy errors. I was able to remove some, but not all.
* tox gives lint errors. Have not figured out how to see what causes the errors.
* tox gives griffe error, ast.BoolOp, but did so even before I edited code
* not sure if Im using tox correctly. I simply run 'tox'

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,23 @@ python3 -m pip install -e .
 'an hour ago'
 ```
 
+#### Also accepts iso8601 text strings
+
+```pycon
+>>> import humanize
+>>> humanize.naturaldate('2007-06-05')
+'Jun 05 2007'
+>>> humanize.naturaldate('2007-06-05T22:00:10')
+'Jun 05 2007'
+>>> humanize.naturaldate('2007-06-05T22:00:10+02:00')
+'Jun 05 2007'
+>>> humanize.naturaltime('22:00:10') # run at 22:00:00
+'10 seconds from now'
+
+Relies on the native `fromisoformat` function that exists on date, datetime and time objects.
+```
+
+
 ### Precise time delta
 
 ```pycon
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -47,6 +47,47 @@ def _now() -> dt.datetime:
     return dt.datetime.now()
 
 
+def _parseiso(
+    value: dt.time | dt.date | dt.datetime | float | str,
+) -> dt.time | dt.date | dt.datetime | float | str:
+    """If string attempts to parse as iso8601 into date, datetime or time.
+
+    Args:
+        value (str or any object): String to be parsed.
+
+    Returns:
+        str (str or any): If string, it will attempt parsing as
+        datetime.date, datetime.datetime or datetime.time, in that order,
+        and return the result of the first succesful parsing, if any.
+
+        Parsing is attempted with the iso8601 function for each of the classes.
+
+        If `value` is not a string or if all attempts at parsing fail,
+        `value` is returned as is.
+
+    """
+    if isinstance(value, str):
+        try:
+            # catches eg '2023-04-01'
+            return dt.date.fromisoformat(value)
+        except ValueError:
+            pass
+
+        try:
+            # catches eg '2023-04-01T12:00:00.123456+02:00', '2023-04-20 12:00:00'
+            return dt.datetime.fromisoformat(value)
+        except ValueError:
+            pass
+
+        try:
+            # catches eg '20:00', '20:00:16.123456', 'T20:00Z'
+            return dt.time.fromisoformat(value)
+        except ValueError:
+            pass
+
+    return value
+
+
 def _abs_timedelta(delta: dt.timedelta) -> dt.timedelta:
     """Return an "absolute" value for a timedelta, always representing a time distance.
 
@@ -246,6 +287,7 @@ def naturaltime(
         str: A natural representation of the input in a resolution that makes sense.
     """
     now = when or _now()
+    value = _parseiso(value)
 
     if isinstance(value, dt.datetime) and value.tzinfo is not None:
         value = dt.datetime.fromtimestamp(value.timestamp())
@@ -274,6 +316,8 @@ def naturalday(value: dt.date | dt.datetime, format: str = "%b %d") -> str:
     formatted according to `format`.
 
     """
+    value = _parseiso(value)
+
     try:
         value = dt.date(value.year, value.month, value.day)
     except AttributeError:
@@ -298,6 +342,8 @@ def naturalday(value: dt.date | dt.datetime, format: str = "%b %d") -> str:
 
 def naturaldate(value: dt.date | dt.datetime) -> str:
     """Like `naturalday`, but append a year for dates more than ~five months away."""
+    value = _parseiso(value)
+
     try:
         value = dt.date(value.year, value.month, value.day)
     except AttributeError:


README.md
```diff
@@ -103,6 +103,23 @@ python3 -m pip install -e .
...
+'Jun 05 2007'
+>>> humanize.naturaldate('2007-06-05T22:00:10')
+'Jun 05 2007'
+>>> humanize.naturaldate('2007-06-05T22:00:10+02:00')
+'Jun 05 2007'
+>>> humanize.naturaltime('22:00:10') # run at 22:00:00
+'10 seconds from now'
+
+Relies on the native `fromisoformat` function that exists on date, datetime and time objects.
+```
```
hugovk, line 120: Please can you move the final triple backticks after '10 seconds from now'?

They indicate where the special code formatting ends.

For example, right now it looks like:

image

https://github.com/velle/humanize/blob/iso8601/README.md#also-accepts-iso8601-text-strings

hugovk (commented):
Thank you for the PR!

> * time.py, added new function _parseiso(value)* relevant functions now call value = _parseiso(value)

👍 

> * wrote inline documentation, and short subparagraph to README.md

👍 

> * have not tested code in any way except the existing pytests which all succeeded

Good the existing tests pass :)

Please can you also add tests? Edit `test_time.py` and for each function we're adding the extra function call, find the corresponding `test_*` test function. Then each one has a list of test input data inside `@pytest.mark.parametrize()`. Let's add some example ISO strings to cover the three different `try`/`except` blocks in the new function.

> * tox gives mypy errors. I was able to remove some, but not all.
> * tox gives lint errors. Have not figured out how to see what causes the errors.

We're only getting mypy errors from lint. Let's come back to these later.

> * tox gives griffe error, ast.BoolOp, but did so even before I edited code

Yep, we can ignore these, it's the docs build and we get the same on `main` and it's not actually failing it:

```
docs: commands[0]> mkdocs build
INFO     -  mkdocstrings: DEPRECATION: mkdocstrings' watch feature is deprecated in favor of MkDocs' watch feature, see https://www.mkdocs.org/user-guide/configuration/#watch
INFO     -  Cleaning site directory
INFO     -  Building documentation to directory: /home/runner/work/humanize/humanize/site
ERROR    -  griffe: .tox/docs/lib/python3.11/site-packages/humanize/time.py:1: Failed to get annotation expression from Tuple: <class 'ast.BoolOp'>
ERROR    -  griffe: .tox/docs/lib/python3.11/site-packages/humanize/time.py:1: Failed to get annotation expression from Tuple: <class 'ast.BoolOp'>
INFO     -  Documentation built in 1.06 seconds
.pkg: _exit> python /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True hatchling.build
  docs: OK (24.70=setup[[23](https://github.com/python-humanize/humanize/actions/runs/5468468419/jobs/9956152679#step:5:24).20]+cmd[1.50] seconds)
  congratulations :) ([24](https://github.com/python-humanize/humanize/actions/runs/5468468419/jobs/9956152679#step:5:25).79 seconds)
```

I'll check them some other time.

> * not sure if Im using tox correctly. I simply run 'tox'

Yep, `tox` is the usual way to run it.

To run a single thing with tox, you can do `tox -e py311` for example (where `e` = tox environment) to only run tests on Python 3.11, or `tox -e lint` to run lint, including mypy.

And `tox -p auto` runs all in parallel.

velle:
Hi Hugo. First of all, thanks for the feedback and for helping me learn this :) I will get back to the other points in your feedback later, but right now I need to figure out how to handle what is described below. 

I started writing tests (some hours ago), and was almost done. But then I realized that humanize does not support datetime.time at all. I had written _parse_iso to also parse into datetime.time, when possible. 

Now I have removed everything that has to do with datetime.time, and edited the documentation accordingly. 

And I have committed and pushed this. This commit only removes the irrelevant time code. It does not include the tests I wrote. Since I imagined that you had to evaluate my commits, before ... merging? ... accepting? them into this iso8601 branch (or into this pull request?). The essence is: I decided to make this commit so it only concerns the removal of datetime.time. 

Should I do more about this commit? Will this commit (and future commits) automatically be part of this pull request, until this pull request is closed or merged?

hugovk:
This PR isn't accepted or merged yet (see the green "Open" at the top). 

Yes, you can commit more things to your branch, and when you push them, they will automatically be part of this PR.

velle:
Hi Hugo. As you noticed, I have been gone, first because of issues with my laptop, and then a week of holiday. But I'm back :)

I just made three commits: 1) Changed every call for _parseiso to _parse_iso. 2) Added tests for _parse_iso() in test_time.py. 3) And then added a few more tests to a few relevant test_x functions in test_time.py. 

Sincerely.

hugovk:
Thank you for the update. Please could you check the lint failures? They're mostly type hint things. Let me know if you'd like a hand.

```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

tests/test_time.py:102:32: F821 undefined name 'Object'
tests/test_time.py:102:50: F821 undefined name 'Object'
...
- hook id: mypy
- exit code: 1

src/humanize/time.py:281: error: Incompatible types in assignment (expression has type "Union[date, float, str]", variable has type "Union[datetime, timedelta, float, str]")  [assignment]
src/humanize/time.py:281: error: Argument 1 to "_parse_iso" has incompatible type "Union[datetime, timedelta, float, str]"; expected "Union[date, datetime, float, str]"  [arg-type]
src/humanize/time.py:310: error: Incompatible types in assignment (expression has type "Union[date, float, str]", variable has type "Union[date, datetime]")  [assignment]
src/humanize/time.py:336: error: Incompatible types in assignment (expression has type "Union[date, float, str]", variable has type "Union[date, datetime]")  [assignment]
tests/test_time.py:102: error: Name "Object" is not defined  [name-defined]
Found 5 errors in 2 files (checked 10 source files)
```

velle:
I managed to make all mypy errors disappear. But it was not easy to do this in an elegant way, when _parse_iso takes and returns a number of different types. The code I have pushed is the most elegant solution I could find, but it does contain the following check/cleanup in naturaltime:

```
_value = _parse_iso(value)
if type(_value) != dt.date:
    # naturaltime is not built to handle dt.date
    value = typing.cast(typing.Union[dt.datetime, dt.timedelta, float, str], _value)
```
 
One solution is modifying `naturaltime` so it would also accept `dt.date` as value. I looked into that but I wasnt sure what the intended behavior should be and if it made sense for all situations and values. 

Another solution is going back to my original approach where the call for fromisoformat() is placed inside each of the naturalX-functions. 

```
def naturaltime(value: dt.datetime | dt.timedelta | float | str)
    if isinstance(value,str):
        try:    value = dt.datetime.fromisoformat(value)
        except: pass

def naturalday(value: dt.date | dt.datetime):
    if isinstance(value,str):
        try:    value = dt.datetime.fromisoformat(value)
        except: pass
        try:    value = dt.date.fromisoformat(value)
        except: pass

def naturaldate(value: dt.date | dt.datetime):
    if isinstance(value,str):
        try:    value = dt.datetime.fromisoformat(value)
        except: pass
        try:    value = dt.date.fromisoformat(value)
        except: pass
```

Or yet another solution is to create two separate helper functions.


```
def _try_parse_datetime(value):
    try:
        return dt.datetime.fromisoformat(value)
    finally:
        return value
    
def _try_parse_date(value):
    try:
        return dt.datetime.fromisoformat(value)
    finally:
        return value

def naturaltime(value: dt.datetime | dt.timedelta | float | str)
    value = _try_parse_datetime(value)
    (...)

def naturalday(value: dt.date | dt.datetime):
    value = _try_parse_date(value)
    value = _try_parse_datetime(value)
    (...)

def naturaldate(value: dt.date | dt.datetime):
    value = _try_parse_date(value)
    value = _try_parse_datetime(value)
    (...)
```

I think I like both of these solutions better than what I pushed. But I'm not sure about anything.

PR #137, opened 2023-08-03
Docs: fix formatting of return value

hugovk:
Currently it's showing each newline as another return value.


hugovk:
# TODO

* [ ] Enable RTD PR previews

PR #147, opened 2023-10-02
Fix DeprecationWarning: Plural value must be an integer, got float

### LINKED ISSUE #106: DeprecationWarning: Plural value must be an integer, got float
### What did you do?

I ran tests.

Tests pass without warnings.

### What actually happened?

The following warning was reported:
```
=============================== warnings summary ===============================
tests/test_i18n.py::test_i18n
  $(@D)/src/humanize/i18n.py:133: DeprecationWarning: Plural value must be an integer, got float
    return get_translation().ngettext(message, plural, num)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

* OS: OpenIndiana
* Python: 3.9.16
* Humanize: 4.6.0

dansarpong:
Fixes #106 

* Ensures plural value gotten is an integer

### DIFF
--- a/src/humanize/i18n.py
+++ b/src/humanize/i18n.py
@@ -123,7 +123,7 @@ def _ngettext(message: str, plural: str, num: int) -> str:
     Returns:
         str: Translated text.
     """
-    return get_translation().ngettext(message, plural, num)
+    return get_translation().ngettext(message, plural, int(num))
 
 
 def _gettext_noop(message: str) -> str:
--- a/src/humanize/i18n.py
+++ b/src/humanize/i18n.py
@@ -123,7 +123,7 @@ def _ngettext(message: str, plural: str, num: int) -> str:
     Returns:
         str: Translated text.
     """
-    return get_translation().ngettext(message, plural, int(num))
+    return get_translation().ngettext(message, plural, num)
 
 
 def _gettext_noop(message: str) -> str:
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -590,7 +590,8 @@ def precisedelta(
     for unit, fmt in zip(reversed(Unit), fmts):
         singular_txt, plural_txt, fmt_value = fmt
         if fmt_value > 0 or (not texts and unit == min_unit):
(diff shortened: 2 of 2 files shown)


src/humanize/i18n.py
```diff
@@ -123,7 +123,7 @@ def _ngettext(message: str, plural: str, num: int) -> str:
     Returns:
         str: Translated text.
     """
-    return get_translation().ngettext(message, plural, num)
+    return get_translation().ngettext(message, plural, int(num))
```
hugovk, line 126: Let's skip this, there may be some cases where we don't want to cast to float (re: https://code.djangoproject.com/ticket/27486). Let's just handle `precisedelta` in this PR, and if the warning comes up for other functions, we can deal with those as appropriate.

```suggestion
    return get_translation().ngettext(message, plural, num)
```

src/humanize/time.py
```diff
@@ -590,7 +590,8 @@ def precisedelta(
     for unit, fmt in zip(reversed(Unit), fmts):
         singular_txt, plural_txt, fmt_value = fmt
         if fmt_value > 0 or (not texts and unit == min_unit):
-            fmt_txt = _ngettext(singular_txt, plural_txt, fmt_value)
+            _fmt_value = 2 if fmt_value > 1 and fmt_value < 2 else int(fmt_value)
+            fmt_txt = _ngettext(singular_txt, plural_txt, _fmt_value)
```
hugovk, line 594: This looks good.

And we can simplify a chained comparison:

```suggestion
            _fmt_value = 2 if 1 < fmt_value < 2 else int(fmt_value)
            fmt_txt = _ngettext(singular_txt, plural_txt, _fmt_value)
```

hugovk (changes requested):
Thanks for the PR! A couple of suggestions.

src/humanize/i18n.py
dansarpong: Sure

src/humanize/time.py
dansarpong: Oh okay, thanks!

hugovk (approved):
Thank you for your contribution!

PR #149, opened 2023-10-06
Add Hebrew localization

clsn:

* Add Hebrew translation .po file
* Uses RIGHT-TO-LEFT MARK characters to try to keep bidirectionality sensible; may not be a good idea.


hugovk:
Thanks, let's try it out and we can make fixes if needed for bidirectionality.

PR #155, opened 2023-11-21
Add Norwegian localization

fdanielsen:
Adds Norwegian ("bokmål") localization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ human-readable size or throughput. It is localized to:
 - Japanese
 - Klingon
 - Korean
+- Norwegian
 - Persian
 - Polish
 - Russian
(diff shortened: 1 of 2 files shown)


hugovk:
Takk!

fdanielsen:
> Takk!

Bare hyggelig! :)

Thanks for the extremely quick response. Looking forward to new release, but also thankful for the support for custom locale directory so we can make use of this even before the official release.

hugovk:
It's about time we did a release, I was hoping https://github.com/pypi/support/issues/2916 would be resolved soon so I could set up release automation, but they have a huge backlog and it's taking much longer, so I'll do one soon. 👍

PR #159, opened 2023-12-03
`metric`: Fix precision format

e-nikolov:
Currently if I try to print a 2 digit number with precision 0, I get this error:
`ValueError: Format specifier missing precision` because the format results in a negative precision number.

This PR makes sure that the precision is at least 0


hugovk:
Please can you share a snippet of code that reproduces the problem, and also include tests?

hugovk:
Thank you!

moi90:
I ran into the same problem (and found the same fix). It would be great to release a new version to PyPI that includes this fix.

hugovk:
Thanks for the ping, we're long overdue a release, I'll make one this week.

hugovk:
4.10.0 is released!

moi90:
Great!

PR #162, opened 2023-12-28
Reword the French translation of very small amount of time

### LINKED ISSUE #161: The French translation of "a moment" should be "un instant"
### What did you do?

The French translation of "a moment" isn't "un moment" but "un instant".

```python
>>> import humanize.i18n
>>> import humanize.time
>>> _ = humanize.i18n.activate("fr_FR")
>>> humanize.time.naturaldelta(0.4)
'un moment'
```

The result should be "un instant".

"un moment" in french is a small but has the property to be actually quite perceptible. It is also not precise. It can mean many seconds, 1minute or even sometimes 5days but is not used for very small amount of time like under 1 second.

The word "moment" actually means that you feel that time but it doesn't look that long. 

"un instant" actually means a very small amount of time. 

```python
>>> import humanize.i18n
>>> import humanize.time
>>> _ = humanize.i18n.activate("fr_FR")
>>> humanize.time.naturaldelta(0.4)
'un instant'
```

### What actually happened?

The result is currently  "un moment"

* OS: Arch
* Python: 3.11.6
* Humanize: 4.8.0

arobase-che:
Fixes #161 

* To reword the french translation of a very small amount of time.

### DIFF
--- a/src/humanize/locale/fr_FR/LC_MESSAGES/humanize.po
+++ b/src/humanize/locale/fr_FR/LC_MESSAGES/humanize.po
@@ -257,7 +257,7 @@ msgstr[1] "%d millisecondes"
 
 #: src/humanize/time.py:164 src/humanize/time.py:259
 msgid "a moment"
-msgstr "un moment"
+msgstr "un instant"
 
 #: src/humanize/time.py:167
 msgid "a second"


hugovk:
Merci!

PR #169, opened 2024-02-02
Fix naturaldelta sub second precision

### LINKED ISSUE #154: naturaldelta failing with float input and sub-second precision
```python
import datetime

import humanize

print(humanize.naturaldelta(datetime.timedelta(seconds=5), minimum_unit="milliseconds"))
>> 5 seconds
print(
    humanize.naturaldelta(datetime.timedelta(seconds=0.5), minimum_unit="milliseconds")
)
>> 500 milliseconds
print(humanize.naturaldelta(5, minimum_unit="milliseconds"))
>> 5 seconds
print(humanize.naturaldelta(0.5, minimum_unit="milliseconds"))
>> 0 milliseconds # !! expected "500 milliseconds"
```

### Versions

* OS: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
* Python: Python 3.11.3
* Humanize: 4.8.0

wolf1986:
Fixes #154 

* Convert to float and not to int
* Explicitly drop support from 'inf' & 'nan' that are supported by float
* In tests, assert both float & timedelta variants are equivalent

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -104,7 +104,7 @@ def naturaldelta(
     Returns:
         str (str or `value`): A natural representation of the amount of time
             elapsed unless `value` is not datetime.timedelta or cannot be
-            converted to int (cannot be float due to 'inf' or 'nan).
+            converted to int (cannot be float due to 'inf' or 'nan').
             In that case, a `value` is returned unchanged.
 
     Raises:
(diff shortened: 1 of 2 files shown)


src/humanize/time.py
```diff
@@ -104,7 +104,8 @@ def naturaldelta(
     Returns:
         str (str or `value`): A natural representation of the amount of time
             elapsed unless `value` is not datetime.timedelta or cannot be
-            converted to int. In that case, a `value` is returned unchanged.
+            converted to int (cannot be float due to 'inf' or 'nan).
```
hugovk, line 107:
```suggestion
            converted to int (cannot be float due to 'inf' or 'nan').
```

hugovk:
Thanks!

PR #188, opened 2024-05-28
Remove shebang from filesize.py

mikelolasagasti:
In #141 shebangs for `number.py` and `time.py` were removed, but `filesystem.py` maintained it.


hugovk:
Thank you!

PR #202, opened 2024-09-09
Drop support for Python 3.8

hugovk:
Python 3.8 is end-of-life next month:

* https://devguide.python.org/versions/
* https://peps.python.org/pep-0569/#lifespan

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


glensc:
It's not dead yet. IMHO Libraries shouldn't be that aggressive with version dropping...

hugovk:
Yep, the 3.8 EOL date is same as 3.13 release date.

Some projects follow SPEC 0 to drop Python versions even earlier. For example, they will drop 3.10 this quarter:

https://scientific-python.org/specs/spec-0000/

PR #222, opened 2024-12-14
Handle en_GB and en_US in i18n

2br-2b:
Fixes #210

Changes proposed in this pull request:

* Add support for en_GB and en_US

PR #230, opened 2025-02-02
Handle en_GB and en_US locale

### LINKED ISSUE #210: Add en_GB and en_US
### What did you do?

Hello! Thank you so much for making this library!!!

I'm using this to bot to translate some strings for my python discord bot. To do that, I am mapping discord locales to Humanize strings, like so:

```python
HUMANIZE_LOCALE_STRINGS = {
    Locale.taiwan_chinese : "zh_HK",
    Locale.german : "de_DE",
    Locale.latin_american_spanish : "es_ES",
    etc.
]
```

I wish I could also map British English and American English to `en_US` and `en_GB`  instead of setting them to `None`, then doing a None check whenever I get the locale.

Especially since `en_GB` is listed as an example locale at the bottom of the Readme, I wish this was an option

---

```python
>>> import humanize
>>> humanize.i18n.activate("en_US")
  File "<stdin>", line 1, in <module>
  File "/home/homer/StoryBot-Testing/.venv/lib/python3.10/site-packages/humanize/i18n.py", line 76, in activate
    translation = gettext_module.translation("humanize", path, [locale])
  File "/usr/lib/python3.10/gettext.py", line 607, in translation
    raise FileNotFoundError(ENOENT,
FileNotFoundError: [Errno 2] No translation file found for domain: 'humanize'
```

dangillet:
Fixes #210 
Replaces existing PR and closes #222 

I tried to push that PR over the finishing line. The commits could all be squashed in a single commit. I wanted to keep track of the original author contribution.

* Add i18n support for en_GB, en_US and any languages starting with en_*


dangillet:
The current type hint for `activate` is
```python
def activate(
    locale: str, path: str | os.PathLike[str] | None = None
) -> gettext_module.NullTranslations:
```

So  `humanize.i18n.activate(None)` is not allowed from a type hinting point of view.

If this is desired though, it's absolutely OK to amend the type hint. I will then also need to slightly amend the function implementation to deal with `None` for `locale`, which would be the same as choosing an English language.

I don't really see the reason for `humanize.i18n.activate(None)`. If you don't want any translation, or because you're done with one language, I would assume the client would use instead `humanize.i18n.deactivate().`

I'll let you decide what you think is best.

hugovk:
Yes, these are good points.

I do see a couple of uses of `humanize.i18n.activate(None if lang == "en" else lang)` in the wild, which admittedly isn't much for a package downloaded 13 million times per month.

https://github.com/search?q=%2Factivate%5C%28None%2F&ref=opensearch&type=code

But we don't know how much non-GitHub code might do `activate(None)`, and on balance I think it's probably better to allow it for compatability. It doesn't cost us much to allow `None`.

What do you think?

dangillet:
I fully agree. It seems like a good idea to allow `activate(None)`. I just wanted to make sure you were aware of it.

I made the change and I moved to test inside the `TestActivate` class as it seems to belong there - we're testing more usage to `activate`.

I then noticed that the original test had a small issue. It did not use the decorator `@freeze_time("2020-02-02")`, which means that when it was calling `humanize.naturaltime(three_seconds)`, the result actually was `5 years ago` which is pretty surprising.

I decided to reduce this _gotcha_ effect by making `NOW` a pytest fixture. Any code that would use it would have its time frozen in `2020-02-02`.

dangillet:
My unit tests were failing on Windows because `gettext` was not installed. I amended `test.yml` file to install `gettext` on Windows and compile the translation files.

The unit tests now pass on Windows but I'm getting two failures for linting and code coverage.

I don't understand the linting issue. Locally I'm not getting any linting issue. I also tried to copy-paste `test.yml` content in https://rhysd.github.io/actionlint/ and I'm not getting any failures.

For code covergage, I'm looking on Sentry, but the only lines that are not tested seem to be the docstring I added to the pytest fixture. What am I missing?

hugovk:
Let's not bother with the .`ps1` scripts, we only need to run it during the automated release process, which is done on Ubuntu. It's one less thing to maintain and test.

(And I've noticed it was accidentally left out when adding the automation, I've reported this in https://github.com/python-humanize/humanize/issues/231.)

dangillet:
Hi,

I'm happy to remove the `ps1` scripts. But the unittests in the CI would not work then for Windows. One of my test uses a translation, and thus needs those files present.

Unless you want me to also change the test so we never try to translate anything in the unittests?

I noticed in other tests this pattern
```python
    try:
        humanize.i18n.activate("ru_RU")
        assert humanize.naturaltime(three_seconds) == "3 секунды назад"
        assert humanize.ordinal(5) == "5ый"
        assert humanize.precisedelta(one_min_three_seconds) == "1 минута и 7 секунд"

    except FileNotFoundError:
        pytest.skip("Generate .mo with scripts/generate-translation-binaries.sh")
```

Is it how we want to handle it? No unittests should run translations on the CI?

hugovk:
Yes, let's use the `try`/`except`.

We can see skips for the Windows CI:

* https://github.com/python-humanize/humanize/actions/runs/13268474110/job/37041909246#step:8:44

But they're run on the macOS and Ubuntu CI:

* https://github.com/python-humanize/humanize/actions/runs/13268474110/job/37041910044#step:8:44
* https://github.com/python-humanize/humanize/actions/runs/13268474110/job/37041910678#step:8:45

Because we're installing `gettext` and running/testing the script for those:

https://github.com/python-humanize/humanize/blob/de8b8a124c5209918ac0f2921d7cf58b1f6c1060/.github/workflows/test.yml#L28-L43

This means people can also test locally without generating the binaries.

tests/test_i18n.py
```diff
@@ -4,18 +4,32 @@
...
 import pytest
 from freezegun import freeze_time
 
 import humanize
 
-with freeze_time("2020-02-02"):
-    NOW = dt.datetime.now()
 
+@pytest.fixture
+def NOW() -> Generator[dt.datetime, None, None]:
```
hugovk, line 16: Instead of this fixture, can we do it similar to `test_time.py` and start the file something like?

```python
with freeze_time("2020-02-02"):
    NOW = dt.datetime.now()
```

Then we also don't need to pass in the fixture below.

hugovk (commented):
Thanks for your patience, I think we're almost there!

tests/test_i18n.py
dangillet: No problem at all! I'm sorry it takes a lot of your time as well.

I certainly can do that. I just want to point out that I came up with that approach because the previous unittest that had been written was surprisingly wrong. It looked like this
```python
def test_en_locale() -> None:
    three_seconds = NOW - dt.timedelta(seconds=3)
    
    humanize.i18n.activate(None)
    test_str = humanize.naturaltime(three_seconds)

    humanize.i18n.activate("en_US")
    assert test_str = humanize.naturaltime(three_seconds)

    humanize.i18n.activate("en_GB")
    assert test_str = humanize.naturaltime(three_seconds)

    humanize.i18n.deactivate()
```

And I simply wanted to check that `test_str == "3 seconds ago"`, but it turned out to be incorrect. The reason is because the test was not decorated with `@freeze_time("2020-02-02")`.

To me this felt a bit like a _gotcha_. Using the global constant `NOW` is not enough. You also need to decorate your test function.

The fixture has the advantage that the time is frozen for the entire test using this fixture. No need to remember to also decorate the test function. But indeed you need to add the fixture to the test function arguments.

I can see that because it's done differently in other test files will create an inconsistency. This might not be desirable.

So I will do the change you suggested. But if you change your mind, it will be easy enough to pop this last commit.

tests/test_i18n.py
```diff
@@ -198,3 +198,35 @@ def test_default_locale_path_undefined__spec__(
...
+        humanize.i18n.activate("en_US")
+        assert test_str == humanize.naturaltime(three_seconds)
+
+        humanize.i18n.activate("en_GB")
+        assert test_str == humanize.naturaltime(three_seconds)
+
+        humanize.i18n.deactivate()
+
+    @freeze_time("2020-02-02")
+    def test_None_locale(self) -> None:
```
hugovk, line 216:
```suggestion
    def test_none_locale(self) -> None:
```

hugovk (approved):
Thank you very much!

tests/test_i18n.py
dangillet: @hugovk 
> I can't actually reproduce this.

To reproduce, consider the following test in `test_i18n.py`:
```python
def test_now() -> None:
    three_seconds = NOW - dt.timedelta(seconds=3)
    assert humanize.naturaltime(three_seconds) == "3 seconds ago" 
```
This results in a failure
```
______________________________________________________________________ test_now _______________________________________________________________________

    def test_now() -> None:
        three_seconds = NOW - dt.timedelta(seconds=3)
    
>       assert humanize.naturaltime(three_seconds) == "3 seconds ago"
E       AssertionError: assert '5 years ago' == '3 seconds ago'
E         
E         - 3 seconds ago
E         + 5 years ago

tests/test_i18n.py:238: AssertionError
=============================================================== short test summary info ===============================================================
FAILED tests/test_i18n.py::test_now - AssertionError: assert '5 years ago' == '3 seconds ago'
  
  - 3 seconds ago
  + 5 years ago

```
The reason is obviously because `humanize.naturaltime` is under the hood calling `dt.datetime.now()` which returns the non-frozen now. The time was frozen for the duration of the context manager when `NOW` has been created and its value _frozen_. But leaving the context manager means that further calls to `now()` will not be patched anymore and we will get the real `now()` value.

Thus this test must be also decorated with `@freeze_time("2020-02-02")`. This could be overlooked, or someone might even set the time wrong.

The option I suggested with the fixture does not suffer from with. The fixture does not leave the context manager `with freeze_time(...)` **for the duration of the test**, and thus any calls to `now()` will use the same frozen time as used for the fixture, which I thought was a way to make writing tests less error prone.

> I think it's fine to put a decorator on the function as well, just to be extra explicit,

We're not being extra explicit. It is **required**. If you remove the decorator from the existing `test_none_locale` it will fail.

I'm sure that clarifies what I meant previously. :smile:

hugovk: That does clarify it, thank you!

Issue #239, opened 2025-02-17
naturalsize output changed between 4.11.0 and 4.12.0


davidboweninrupt:
Possibly a result of https://github.com/python-humanize/humanize/pull/218, we are seeing that the output has changed:

1. podman run -it python bash
2. python -m pip install humanize==4.11.0
3. python -c "import humanize; print(humanize.naturalsize(1.123456789))"
> 1 Bytes
4. python -m pip install humanize==4.12.0
5. python -c "import humanize; print(humanize.naturalsize(1.123456789))"
> 1.123456789 Bytes

Expected behaviour is that output would not change in a minor version or that it would be [documented in the release notes](https://github.com/python-humanize/humanize/releases/tag/4.12.0).

hugovk:
Thanks for the report.

This was caused by some bad formatting replacement in https://github.com/python-humanize/humanize/commit/fb61d5aae902e4526cd733a90d7a96ecded11b0e / https://github.com/python-humanize/humanize/pull/226.

Let's fix it properly and make a patch release.

---

As an aside, we get:

```pycon
>>> import humanize
>>> humanize.naturalsize(1.123456789 * 1000)
'1.1 kB'
>>> humanize.naturalsize(1.123456789 * 1000000)
'1.1 MB'
```

So I think it would be better to also get:

```pycon
>>> humanize.naturalsize(1.123456789 * 1000000)
'1.1 Bytes'
```

Instead of:


```pycon
>>> humanize.naturalsize(1.123456789 * 1000000)
'1 Bytes'
```

But that's another issue. And we're definitely missing tests for non-integer input:

https://github.com/python-humanize/humanize/blob/a0602c70bb6e11322107b51114333bab7adab021/tests/test_filesize.py#L14-L75

hugovk:
https://github.com/python-humanize/humanize/pull/240 adds tests and should revert to the previous behaviour. How does it look?

PR #240, opened 2025-02-17
Fix regression in `naturalsize` for float

hugovk:
Fixes https://github.com/python-humanize/humanize/issues/239.

### DIFF
--- a/src/humanize/filesize.py
+++ b/src/humanize/filesize.py
@@ -94,7 +94,7 @@ def naturalsize(
         return f"{bytes_} Byte"
 
     if abs_bytes < base:
-        return f"{bytes_}B" if gnu else f"{bytes_} Bytes"
+        return f"{int(bytes_)}B" if gnu else f"{int(bytes_)} Bytes"
 
     for i, s in enumerate(suffix, 2):
         unit = base**i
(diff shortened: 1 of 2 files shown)


davidboweninrupt:
Hi @hugovk 

That's actually wonderful.

I can accept either behaviour, but I wanted it to be stable. Your additional tests really help with that. Ironically the change in behaviour popped up because my tests spotted it.

My local fix was also to convert the bytes to an integer.

I do have cases where I'm looking at the average-bytes-per-second so it might not be an integer. However, `naturalsize` does say it's trying to be like a file size so integers do make sense.

The proposed fix does an `int` rather than rounding it and as we know:

> int(2.7)
> 2

which is tolerable, but not what I'd expect from `humanize`. That's mostly because I've been so impressed by how `humanize` has given me great outputs for so long.

So, in summary, I'd be fine with it going in this way.

I really appreciate you getting to this so quickly.

hugovk:
Yeah, and it matches the behaviour of the earlier percent formatting:

```pycon
>>> "%d" % 2.7
'2'
>>> int(2.7)
2
```

I'm open to changing the rounding, but let's do it in a more intentional way than a bad formatting upgrade :)

davidboweninrupt:
Nice, thanks @hugovk

hugovk:
You're welcome! 🚀

PR #243, opened 2025-03-24
Update license file reference from LICENSE to LICENCE in pyproject.toml

williamjamir:
The `humanize` package does not currently include the `LICEN[CS]E` file when uploaded to PyPI, causing tools like `pip-license` to fail to detect that this package is MIT licensed. 

By updating the reference to `LICENCE`, we ensure that the correct file is included in the package.


hugovk:
Thanks for the fix, the test failures are unrelated.

PR #248, opened 2025-04-01
Improvements for French translation

merwok:
* fix number units translations
* use proper decimal separator

### DIFF
--- a/src/humanize/locale/fr_FR/LC_MESSAGES/humanize.po
+++ b/src/humanize/locale/fr_FR/LC_MESSAGES/humanize.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-01-08 19:22+0200\n"
+"POT-Creation-Date: 2025-04-01 12:17-0400\n"
 "PO-Revision-Date: 2013-06-22 08:52+0100\n"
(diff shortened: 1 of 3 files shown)


src/humanize/locale/fr_FR/LC_MESSAGES/humanize.po
```diff
@@ -119,257 +119,247 @@ msgctxt "9 (female)"
 msgid "th"
 msgstr "e"
 
-#: src/humanize/number.py:178
+#: src/humanize/number.py:183
```
merwok, line 122: In my own projects, I tend to save filenames but not line numbers to PO files, to avoid such diff churn. What do you think of doing that change in another PR?

hugovk: I don't make translations myself, so if they're not useful I'll take your word for it and a PR is welcome :)

hugovk:
Merci beuacoup !

src/humanize/locale/fr_FR/LC_MESSAGES/humanize.po
merwok: Either the line numbers get outdated, so they are useless and get ignored, or the create churn in PRs (which I could just accept but…)

So personally I find the filenames useful to be able to look up the context (source strings and even context lines from pgettext are not always enough), I tend to dislike the line numbers.

But you could open a ticket and ping translators to see their experience!

PR #250, opened 2025-04-08
Fix regression in `naturalsize` for float and str

loicleyendecker:
Currently if `gnu` is set to False, calling `naturalsize(1.0)` or `naturalsize("1")` will return `"1.0 Byte` instead of `"1 Byte"`, which was the return value until 4.12.

### DIFF
--- a/tests/test_filesize.py
+++ b/tests/test_filesize.py
@@ -69,6 +69,8 @@ import humanize
         ([3000000, False, True], "2.9M"),
         ([1024, False, True], "1.0K"),
         ([1, False, False], "1 Byte"),
+        ([1.0, False, False], "1 Byte"),
+        (["1", False, False], "1 Byte"),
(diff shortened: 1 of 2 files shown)


tests/test_filesize.py
```diff
@@ -79,6 +79,7 @@
         ([1.123456789, False, True], "1B"),
         ([1.123456789 * 10**3, False, True], "1.1K"),
         ([1.123456789 * 10**6, False, True], "1.1M"),
+        ([1.0, False, False], "1 Byte"),
```
hugovk, line 82: Please could you move this to after line 71, after `[1, False, False], "1 Byte"),`?

Let's also add a test case with "1" as well.

hugovk (commented):
Thanks for the fix.

Like https://github.com/python-humanize/humanize/issues/239, this was also caused by https://github.com/python-humanize/humanize/pull/226 and this is the similar fix for the other case.

PR #295, opened 2026-02-17
Fix `naturaldelta` truncating years instead of rounding

ArmaanjeetSandhu:
* Modified `naturaldelta` in `src/humanize/time.py` to use `round()` instead of integer division (`//`) for year calculations. This fixes a bug where 2.9 years was incorrectly formatted as "2 years" instead of "3 years".
* Added a new test case to `tests/test_time.py` to verify that a timedelta of 1058 days (~2.9 years) correctly rounds to "3 years".
* Ensured consistency with the docstring claim that the timedelta is "rounded to the nearest unit".

### DIFF
--- a/tests/test_time.py
+++ b/tests/test_time.py
@@ -127,6 +127,7 @@ def test_naturaldelta_nomonths(test_input: dt.timedelta, expected: str) -> None:
         (dt.timedelta(days=9), "9 days"),
         (dt.timedelta(days=365), "a year"),
         (dt.timedelta(days=365 * 1_141), "1,141 years"),
+        (dt.timedelta(days=1058), "3 years"),
         ("NaN", "NaN"),  # Returns non-numbers unchanged.
         # largest possible timedelta
         (dt.timedelta(days=999_999_999), "2,739,726 years"),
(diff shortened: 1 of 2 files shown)


tests/test_time.py
```diff
@@ -127,6 +127,7 @@ def test_naturaldelta_nomonths(test_input: dt.timedelta, expected: str) -> None:
         (dt.timedelta(days=9), "9 days"),
         (dt.timedelta(days=365), "a year"),
         (dt.timedelta(days=365 * 1_141), "1,141 years"),
+        (dt.timedelta(days=1058), "3 years"),
```
hugovk: Let's add a few more cases:
```suggestion
        (dt.timedelta(days=365 * 2), "2 years"),
        (dt.timedelta(days=365 * 2.4), "2 years"),
        (dt.timedelta(days=365 * 2.6), "3 years"),
        (dt.timedelta(days=365 * 2.9), "3 years"),
        (dt.timedelta(days=365 * 3), "3 years"),
```

hugovk (approved):
Thanks!

PR #304, opened 2026-03-31
Add test for exact googol (10**100) in `intword()`

### LINKED ISSUE #134: intword fails to parse numbers >= 1 Googol
### What did you do?
I tried to use humanize to convert `10_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000` (1 Googol) to the text output “1 googol” by calling `humanize.intword()` with the number.

I expected the text output to be “1 googol” as the [documentation for intword](https://humanize.readthedocs.io/en/latest/number/#humanize.number.intword) claims the library can handle numbers of this size.

### What actually happened?
The output was just the number as a string, without any word at all: `'10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'`

This is also the case for numbers larger than 1 Googol.

* OS: Windows 11 Pro 22H2 22621.1992
* Python: Python 3.11.4 (also tried 3.10.11, 3.9.13, 3.8.10, and 3.7.9)
* Humanize: 4.7.0 (4.6.0 for Python 3.7.9)

The [best reproductions](https://stackoverflow.com/help/minimal-reproducible-example) are [self-contained scripts](https://ericlippert.com/2014/03/05/how-to-debug-small-programs/) with minimal dependencies.

```python
>>> import humanize
# 1 Googol:
>>> GOOGOL = 10_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000
>>> humanize.intword(GOOGOL)
'10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'

# 1 Googol minus 1:
>>> humanize.intword(GOOGOL - 1)
'10000000000000001323945434466030186557813051577054744406252071157760.0 decillion'

# 1000 Googol:
>>> humanize.intword(GOOGOL * 1000)
'10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'
```

The fact that it doesn't seem to work in any Python versions going all the way back to 3.8 in humanize v4.7.0 (and Python 3.7 with v4.6.0) seems odd to me. It leads me to believe that either I am misunderstanding the documentation, I am using the library wrongly, or something is wrong with how humanize handles a Googol.
If humanize is not supposed to parse a googol to the string “1 googol”, the docs should be updated to reflect that. Currently, they state:
> Supports up to decillion (33 digits) and googol (100 digits).
Which I understood meaning “Can parse and display a googol”.

My apologies if the problem turns out to be me misunderstanding/-using.

BlocksecPHD:
## Summary
This PR adds a test case for exact googol (`10**100`) in `intword()` function.

## Context
- Issue #134 reported that `intword(10**100)` returns the raw number instead of "1.0 googol"
- PR #273 fixed the underlying issue by using `bisect` instead of for-loop
- However, the test for exact googol was not added in that PR
- The existing tests cover `2e100` (2 googols) and `10**101` (10 googols), but not `10**100` (1 googol)

## Changes
Added test case `([10**100], "1.0 googol")` to `tests/test_number.py`

## Testing
The test passes on the current repo code (which already includes the fix from PR #273). This ensures the googol edge case is covered in the test suite for future releases.

Fixes #134

### DIFF
--- a/tests/test_number.py
+++ b/tests/test_number.py
@@ -122,6 +122,7 @@ def test_intword_powers() -> None:
         ([2e100], "2.0 googol"),
         ([None], "None"),
         (["1230000", "%0.2f"], "1.23 million"),
+        ([10**100], "1.0 googol"),
         ([10**101], "10.0 googol"),
         ([math.nan], "NaN"),
         ([math.inf], "+Inf"),


hugovk:
> ## Testing
> This test will pass once version 4.15.0 (containing the fix from PR #273) is released to PyPI.

What do you mean? It has been released: https://pypi.org/project/humanize/4.15.0/

And why does it matter if it's released to PyPI? The tests test the code in this repo.

BlocksecPHD:
Hi @hugovk, thank you for pointing that out! You're right - the tests test the code in this repo, not the PyPI package. 

I've updated the PR description to correct that misleading statement. The test passes on the current repo code (which already has the fix from PR #273), and ensures the googol edge case is covered in the test suite.

PR #308, opened 2026-04-11
Improve performance of `number.scientific`

hugovk:
Replace for loop with list comprehension.

And add perflint to Ruff rules.


hugovk:
Already merged in https://github.com/python-humanize/humanize/pull/313 🚀

PR #316, opened 2026-05-18
Add `conjunction` parameter to `natural_list`

### LINKED ISSUE #214: Proposal for additional `conjunction` parameter in `natural_list` function
Thank you for the new version; `natural_list` really resolves several issues.

### Summary

I would like to propose an enhancement to the `natural_list` function. Currently, the output for the list `[1, 2, 3]` is formatted as "1, 2 and 3". I propose adding a parameter that allows users to choose between "and" or "or" in the output.

> I checked the documentation and found in #103 that there is interest in implementing something like this issue.

### Motivation

This enhancement would provide more flexibility in how lists are presented, particularly in contexts where the choice between "and" and "or" is significant.

### Proposed Changes

1. Introduce a new parameter (e.g., `conjunction`) to the `natural_list` function.
2. Accept boolean values for this parameter:
   - `True` (default behavior)
   - `False`

### Example Usage

With the proposed change, the usage would look like this:

```python
from humanize import natural_list

print(natural_list([1, 2, 3]))  # Output: "1, 2 and 3"
print(natural_list([1, 2, 3], conjunction=False))  # Output: "1, 2 or 3"
```

### Another Idea

We could also consider adding `or` and `and` to the i18n support for multiple languages, but I believe that should be addressed in a separate issue.

---

I could submit the PR if there's interest, but I'm not sure if `conjunction` would be the best name for this parameter.

SAY-5:
Fixes #214

* Add a keyword-only `conjunction` parameter to `natural_list`, defaulting to `"and"` so existing behaviour is unchanged.
* This implements the design hugovk suggested in the issue, allowing any string such as `"or"` or `"and also"`.
* Add tests covering the new parameter alongside the default.

### DIFF
--- a/src/humanize/lists.py
+++ b/src/humanize/lists.py
@@ -9,10 +9,11 @@ if TYPE_CHECKING:
 __all__ = ["natural_list"]
 
 
-def natural_list(items: list[Any]) -> str:
+def natural_list(items: list[Any], *, conjunction: str = "and") -> str:
     """Natural list.
 
-    Convert a list of items into a human-readable string with commas and 'and'.
+    Convert a list of items into a human-readable string with commas and a
+    conjunction.
 
     Examples:
         >>> natural_list(["one", "two", "three"])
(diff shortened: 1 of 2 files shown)


SAY-5:
Any interest in this? It adds a conjunction argument to natural_list so callers can get "or" instead of the hardcoded "and".

PR #319, opened 2026-05-21
chore: improve humanize maintenance path

lphuc2250gma:
Summary:
- Add explicit unit tests covering edge-case inputs (e.g., None, NaN, very large/negative values, non-numeric strings) for intword/intcomma/fractional in src/humanize/number.py, and tighten type annotations on any helper that currently lacks precise return types.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/src/humanize/number.py
+++ b/src/humanize/number.py
@@ -418,13 +418,13 @@ def scientific(value: NumberOrString, precision: int = 2) -> str:
 
 
 def clamp(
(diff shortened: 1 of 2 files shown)


hugovk:
https://github.com/hugovk/norwegianblue/pull/300

PR #320, opened 2026-05-24
Stop printing two minus signs in fractional for a negative mixed number

c-tonneslan:
\`fractional(-1.3)\` currently returns \`'-1 -3/10'\`. Same shape for any negative number whose absolute value is at least 1: \`fractional(-2.5)\` returns \`'-2 -1/2'\`, etc.

\`int()\` truncates toward zero, so \`int(-1.3)\` is \`-1\` and \`-1.3 - (-1)\` is \`-0.3\`, which \`Fraction\` keeps as \`-3/10\`. Both the whole-number part and the numerator end up signed and the format string prints them both verbatim.

The minus sign already rides on the whole-number part, so absorb it from the numerator with \`abs()\`. The result is the conventional mixed-fraction form \`'-1 3/10'\` (= -1.3).

Added negative-input cases to \`test_fractional\`.

### DIFF
--- a/src/humanize/number.py
+++ b/src/humanize/number.py
@@ -367,7 +367,12 @@ def fractional(value: NumberOrString) -> str:
     if not whole_number:
         return f"{numerator:.0f}/{denominator:.0f}"
 
-    return f"{whole_number:.0f} {numerator:.0f}/{denominator:.0f}"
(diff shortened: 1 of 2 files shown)


hugovk:
Thank you!

PR #321, opened 2026-05-28
Fix ordinal() suffix for negative integers

fbindakheel:
## Bug

`ordinal()` returns the wrong suffix for negative integers:

```pycon
>>> humanize.ordinal(-1)
'-1th'   # expected '-1st'
>>> humanize.ordinal(-2)
'-2th'   # expected '-2nd'
>>> humanize.ordinal(-21)
'-21th'  # expected '-21st'
```

The docstring states the function "works for any integer," so negatives are in scope.

## Root cause

The suffix is selected with `value % 10` / `value % 100`. Python's modulo on negative numbers returns a non-negative result (`-1 % 10 == 9`), so negative values index into the wrong `_ORDINAL_SUFFIXES` slot (`9` → `"th"`). Some negatives (e.g. `-11`, `-12`) happen to land on `"th"` correctly by coincidence, but `-1`, `-2`, `-3`, `-21`, `-101`, etc. are wrong.

## Fix

Compute the suffix from `abs(value)`. The displayed number keeps its sign. One-line logic change.

## Tests

Added negative cases to the `test_ordinal` parametrization (`-1`→`-1st`, `-2`→`-2nd`, `-3`→`-3rd`, `-11`→`-11th`, `-21`→`-21st`, `-101`→`-101st`, `-111`→`-111th`, etc.). Verified the new cases fail on the current code and pass with the fix; full suite (696 tests) passes.


hugovk:
> The docstring states the function "works for any integer," so negatives are in scope.

Strictly speaking maybe, but perhaps the original intent was for "any positive integer"?

Stepping back a moment: do negative ordinals like -1st, -2nd or -21st even make sense?

hugovk:
Closing due to lack of reply, and the fact that negative ordinals don't actually appear to be a thing in any language.

@fbindakheel Please don't waste maintainers' time by opening PRs and then not replying to questions.

PR #326, opened 2026-06-11
Fix doubled negative sign in fractional() for negative mixed numbers

patchwright:

* `fractional()` produced a doubled negative sign for negative mixed numbers. `fractional(-1.3)` returned `'-1 -3/10'` instead of `'-1 3/10'`, and `fractional(-2.5)` returned `'-2 -1/2'`. The whole-number part already carries the sign (`int(-1.3) == -1`), so repeating it on the numerator is wrong. This is inconsistent with the pure-fraction case, which is already correct (`fractional(-0.3) == '-3/10'`).
* Fix: use `abs(numerator)` in the mixed-fraction output. Positives and pure fractions are unaffected.
* Add regression cases (`-0.3`, `-1.3`, `-2.5`) to the parametrized `test_fractional` — it previously had no negative inputs. Verified the new cases fail on the current code and pass with the fix; the full `tests/test_number.py` suite (222 tests) passes and `ruff` is clean.

No existing issue.


hugovk:
Duplicate of https://github.com/python-humanize/humanize/pull/320.

PR #327, opened 2026-06-12
Fix i18n locale activation

VladislavLysykh:
Added tests for i18n locale activation.

Changes:
- added coverage for locale activation
- verified translation loading
- verified locale switching behavior

Test results:
703 passed, 69 skipped


hugovk:
How is this testing locale activation? It's only testing some noop functions that don't really do anything.

hugovk:
@VladislavLysykh Please don't waste time with stuff like this.

PR #328, opened 2026-06-16
Carry `metric()` to the next SI prefix when rounding reaches 1000

gaoflow:
## Summary

`metric()` can emit a four-digit mantissa with a too-small SI prefix:

```python
>>> humanize.metric(999.9, "V")
'1000 V'        # expected '1.00 kV'
>>> humanize.metric(999.99, "V")
'1000 V'        # expected '1.00 kV'
>>> humanize.metric(999999, "V")
'1000 kV'       # expected '1.00 MV'
>>> humanize.metric(0.0009999, "V")
'1000 μV'       # expected '1.00 mV'
```

This contradicts the function's own docstring, which states the prefix is

> always chosen so that non-significant zero digits are required ... `1,230,000` will become `1.23M` instead of `1230K`

`1000 V` / `1000 kV` are precisely the `1230K`-style outputs it promises to avoid, so the documented contract is the oracle here.

## Root cause

`metric()` selects the SI prefix bucket from `exponent = floor(log10(abs(value)))` on the raw value, then rounds the mantissa to the requested precision. When the value sits just below a power-of-1000 boundary, the rounded mantissa becomes `1000`, but the prefix was already locked in from the pre-rounding exponent, and the code never re-checks whether rounding crossed into the next bucket.

## Fix

After computing the mantissa and its display precision, if `round(abs(mantissa), digits) >= 1000` and a higher prefix is available (`exponent < 30`), advance to the next prefix bucket and re-divide. The top of the SI range (quetta, `Q`, 10^30) has no higher prefix, so the `exponent < 30` guard leaves it unchanged rather than indexing past the prefix table.

## Tests

Added four cases to the `test_metric` parametrization (the carry cases above). They fail on `main` (`'1000 V' != '1.00 kV'`, etc.) and pass with this change. The full `tests/test_number.py` suite passes (223 passed), so existing `metric()` output, including the high-range `QF`/`qA` cases and the exact power-of-1000 inputs, is unchanged.

Disclosure: I developed this fix with AI assistance, under my direction. I reviewed and verified the boundary behaviour (positive carry, the negative-exponent `mV` path, and the top-of-range quetta guard) and the regression run myself, and I am happy to adjust the approach.


src/humanize/number.py
```diff
@@ -538,14 +538,21 @@ def metric(value: float, unit: str = "", precision: int = 3) -> str:
...
 
-    value /= 10 ** (exponent // 3 * 3)
+    old_bucket = exponent // 3 * 3
+    value /= 10**old_bucket
+    digits = int(max(0, precision - exponent % 3 - 1))
+    if exponent < 30 and round(abs(value), digits) >= 1000:
+        exponent += 3 - exponent % 3
+        new_bucket = exponent // 3 * 3
+        value /= 10 ** (new_bucket - old_bucket)
+        digits = int(max(0, precision - exponent % 3 - 1))
```
hugovk, line 548:
```suggestion
    if exponent < 30 and round(abs(value), digits) >= 1000:
        exponent += 3 - exponent % 3
        new_bucket = exponent // 3 * 3
        value /= 10 ** (new_bucket - old_bucket)
        digits = int(max(0, precision - exponent % 3 - 1))

```

hugovk (approved):
Thanks!

PR #331, opened 2026-06-24
Support Romanian ordinal formatting

### LINKED ISSUE #270: Impossible to translate to Romanian
I was checking out the package to use it in a project. I'd need to translate it to Romanian, however a summary look reveals that would be a challenging task. For example the ordinal seems to make a map of suffixes for numbers 0-9, but that doesn't work in Romanian:

<dl>
    <dt>0th</dt>
    <dd>to my knowledge doesn't exist, but maaaybe among friends you could say "al 0-lea" (m); maybe; no (f) though;</dd>
    <dt>1st</dt>
    <dd>"prima" (f); "primul" (m);</dd>
    <dt>2nd</dt>
    <dd>"a 2-a" (f); "al 2-lea" (m);</dd>
    <dt>3rd</dt>
    <dd>"a 3-a" (f); "al 3-lea" (m);</dd>
    <dt>23rd</dt>
    <dd>"a 23-a" (f); "al 23-lea" (m).</dd>
</dl>

So to give it a chance to translate at least the possible strings, they would need to interpolate the value.

lin-hongkuan:
Fixes #270

* Allow ordinal translations to provide a complete value-aware format while preserving the existing suffix-only fallback.
* Add Romanian ordinal translations for male and female forms, including the special `1` forms and numeric forms such as `21`/`23`.
* Add regression coverage for Romanian ordinal output.

Validation:

* `bash scripts/generate-translation-binaries.sh`
* `python -m pytest -q` (`776 passed`)
* `python -m ruff check .`
* `python -m ruff format --check .`

### DIFF
--- a/src/humanize/locale/ro/LC_MESSAGES/humanize.po
+++ b/src/humanize/locale/ro/LC_MESSAGES/humanize.po
@@ -1,0 +1,126 @@
+# Romanian translations for humanize.
+# This file is distributed under the same license as the humanize package.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: humanize\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2026-06-24 00:00+0000\n"
+"PO-Revision-Date: 2026-06-24 00:00+0000\n"
+"Last-Translator: Automatically generated\n"
+"Language-Team: Romanian\n"
+"Language: ro\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : (n==0 || (n%100 > 0 && n%100 < 20)) ? 1 : 2);\n"
+
+#: src/humanize/number.py
+msgctxt "ordinal value 1 (male)"
+msgid "%(value)s"
+msgstr "primul"
+
+#: src/humanize/number.py
+msgctxt "ordinal value 1 (female)"
+msgid "%(value)s"
+msgstr "prima"
+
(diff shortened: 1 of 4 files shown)


lin-hongkuan:
I do not have permission to add labels on this repository. This should use the `changelog: Fixed` label for the required PR label workflow.

src/humanize/number.py
```diff
@@ -108,7 +109,11 @@ def ordinal(value: NumberOrString, gender: str = "male") -> str:
         return str(value)
     gender = "male" if gender == "male" else "female"
     digit = 0 if value % 100 in (11, 12, 13) else value % 10
-    return f"{value}{P_(f'{digit} ({gender})', _ORDINAL_SUFFIXES[digit])}"
+    ordinal_format = P_(f"ordinal value {value} ({gender})", _ORDINAL_VALUE)
+    if ordinal_format == _ORDINAL_VALUE:
+        suffix = P_(f"{digit} ({gender})", _ORDINAL_SUFFIXES[digit])
+        ordinal_format = P_(f"ordinal {digit} ({gender})", f"{_ORDINAL_VALUE}{suffix}")
+    return ordinal_format % {"value": value}
```
hugovk, line 116: Could you explain a bit how this works? Any special instructions for the translator?

src/humanize/locale/ro/LC_MESSAGES/humanize.po
hugovk, line 1: This file only translates `src/humanize/number.py`, and not the other files. Is that intentional?

hugovk (commented):
@alexei Please could you check if this works for Romanian?

hugovk:
AI generated, no reply, no indication the author knows or needs Romanian, or that this is correct as it affects all locales.

PR #332, opened 2026-06-27
fix: handle decimal string input in intword()

jtvhd6:
Bug: intword("12400.9") returns "12400.9" instead of "12.4 thousand" because int("12400.9") raises a ValueError in Python.
Fix: Changed value = int(value) to value = int(float(value)) so decimal strings are handled correctly, consistent with how intcomma() already handles this case.


hugovk:
Thanks for the PR, please could you add test cases that failed before and pass with this fix?

For example, `("1234.5", "1.2 thousand")` and `("1234.9", "1.2 thousand")` to `test_intword`.

PR #334, opened 2026-06-29
Return non-finite floats unchanged from `naturaldelta`

### LINKED ISSUE #333: naturaldelta() raises OverflowError on float('inf') instead of returning it unchanged
### Summary

### Reproduction
```python
import humanize
```

### Expected

### Actual
Uncaught `OverflowError`.

### Fix sketch

### Environment

uttam12331:
## Summary
Closes #333.

`naturaldelta()`'s docstring says values it cannot convert to a `timedelta` are returned unchanged, and `float("nan")` already is. But `float("inf")` / `float("-inf")` raised an uncaught `OverflowError` (from `int(value)`) instead of being returned unchanged like `nan`:

```pycon
>>> humanize.naturaldelta(float("nan"))
'nan'
OverflowError: cannot convert float infinity to integer   # before
'inf'                                                      # after
```

## Fix
Catch `OverflowError` and return the value unchanged **only when it is non-finite**. A too-large *finite* value — whose `OverflowError` comes from `timedelta(seconds=...)` — is still re-raised, preserving the documented `Raises: OverflowError` contract:

```python
except OverflowError:
    if not math.isfinite(value):
        return str(value)
    raise
```

## Tests
- `test_naturaldelta_non_finite` — asserts `nan`/`inf`/`-inf` return `"nan"`/`"inf"`/`"-inf"`.
- `test_naturaldelta_too_large_value_raises` — locks in that a too-large finite value (`1e30`) still raises `OverflowError`.

Verified locally: `pytest tests/test_time.py` → **385 passed**; `ruff check` clean.


tests/test_time.py
```diff
@@ -138,6 +138,19 @@ def test_naturaldelta(test_input: float | dt.timedelta, expected: str) -> None:
         assert humanize.naturaldelta(-test_input) == expected
 
 
+def test_naturaldelta_non_finite() -> None:
+    """Non-finite floats are returned unchanged instead of raising (#333)."""
+    assert humanize.naturaldelta(float("nan")) == "nan"
+    assert humanize.naturaldelta(float("inf")) == "inf"
+    assert humanize.naturaldelta(float("-inf")) == "-inf"
```
hugovk, line 145: Let's parameterise:

```suggestion
@pytest.mark.parametrize("value, expected", [
      (float("nan"), "nan"), (float("inf"), "inf"), (float("-inf"), "-inf"),
  ])
def test_naturaldelta_non_finite() -> None:
    """Non-finite floats are returned unchanged instead of raising."""
    assert humanize.naturaldelta(value) == expected
```

src/humanize/time.py
```diff
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import math
```
hugovk: Let's defer this so it's not imported if not needed.

Please move it inside the `except OverflowError:` block.


```suggestion
```

hugovk (approved):
Thanks, just a couple of suggestions.

src/humanize/time.py
```diff
@@ -153,6 +154,14 @@ def naturaldelta(
             delta = dt.timedelta(seconds=value)
         except (ValueError, TypeError):
             return str(value)
+        except OverflowError:
+            # `int(value)` raises OverflowError for non-finite floats (inf/-inf),
+            # which, like NaN, are returned unchanged. A too-large *finite* value
+            # (whose OverflowError comes from `timedelta`) is still raised, per
+            # the documented `OverflowError` contract.
+            if not math.isfinite(value):
```
hugovk, line 163:
```suggestion
            import math

            if not math.isfinite(value):
```

PR #339, opened 2026-07-01
ordinal: suffix negative integers correctly

HrachShah:


hugovk:
Duplicate of https://github.com/python-humanize/humanize/pull/321, but more importantly: https://github.com/python-humanize/humanize/pull/349#issuecomment-4926962834

PR #342, opened 2026-07-02
naturaldelta: include OverflowError in the int() guard so float('inf') returns unchanged

### LINKED ISSUE #333: naturaldelta() raises OverflowError on float('inf') instead of returning it unchanged

HrachShah:
Fixes #333.

`naturaldelta()` documents that a non-finite float value (e.g. `NaN`, `inf`) is "returned unchanged." The previous code wrapped the `int(value)` probe in `except (ValueError, TypeError)`, which catches `NaN` (`int(float('nan'))` raises `ValueError`) but misses `float('inf')` and `float('-inf')` — those raise `OverflowError` from the same conversion. The result was an uncaught `OverflowError: cannot convert float infinity to integer` for `humanize.naturaldelta(float('inf'))`, in contradiction with the docstring.

Adding `OverflowError` to the caught tuple restores the documented behaviour: `naturaldelta(float('inf'))` now returns `'inf'` and `naturaldelta(float('-inf'))` returns `'-inf'`, symmetric with the existing `NaN` case.

Two new parametrized cases (`float("inf")` and `float("-inf")`) on the existing `test_naturaldelta` exercise the issue. They fail on the pre-fix code with the exact `OverflowError` reported in #333 and pass with the fix. The `math.isfinite(test_input)` guard on the existing `naturaldelta(-test_input)` symmetry assertion is needed because `-float('inf')` is `float('-inf')` (and vice versa), so negating and re-formatting would not round-trip.

Verified with `python3 -m pytest tests/`: 727 passed, 69 skipped (unrelated).


hugovk:
There's already an approved PR for this: https://github.com/python-humanize/humanize/pull/334.

What's wrong with that PR? What does this PR add or do better?

hugovk:

PR #343, opened 2026-07-04
Return non-finite floats unchanged from naturaldelta

### LINKED ISSUE #333: naturaldelta() raises OverflowError on float('inf') instead of returning it unchanged

eeshsaxena:
## Summary

Fixes #333.

`naturaldelta()` documents that a value which cannot be converted to a number is "returned unchanged." `float("nan")` is handled (its `int()` raises `ValueError`, which is caught), but `float("inf")`/`float("-inf")` raise an uncaught `OverflowError` from `int(value)`, because the `except (ValueError, TypeError)` clause omitted `OverflowError`. The behavior was asymmetric and contradicted the docstring.

### Before
```python
>>> humanize.naturaldelta(float("nan"))
'nan'
```

### After
```python
'inf'
>>> humanize.naturaldelta(float("-inf"))
'-inf'
```

## Implementation note

The catch is scoped to the `int(value)` guard only. A *finite* value too large to convert to a `datetime.timedelta` (e.g. `1e40`) still raises `OverflowError` from `timedelta()`, as documented under `Raises:` - this is verified by an added test so the non-finite guard does not swallow the legitimate overflow.

## Tests

Added `test_naturaldelta_nonfinite` (nan, inf, -inf returned unchanged) and `test_naturaldelta_overflow` (finite `1e40` still raises `OverflowError`). Full `tests/test_time.py` passes (387 passed).

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -148,11 +148,16 @@ def naturaldelta(
         delta = value
     else:
         try:
-            int(value)  # Explicitly don't support string such as "NaN" or "inf"
-            value = float(value)
-            delta = dt.timedelta(seconds=value)
-        except (ValueError, TypeError):
+            # Explicitly don't support non-finite or non-numeric values such as
+            # "NaN"/"inf": int() raises ValueError for NaN and OverflowError for
+            # infinity, and TypeError for unsupported types.
+            int(value)
+        except (ValueError, TypeError, OverflowError):
             return str(value)
+        value = float(value)
+        # A finite value that is too large for timedelta still raises
+        # OverflowError here, as documented.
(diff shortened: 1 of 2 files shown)


hugovk:
There are now four open PRs to fix #333, and one is already an approved: https://github.com/python-humanize/humanize/pull/334.

eeshsaxena:
You're right - #334 already lands the same fix (catch `OverflowError`, return non-finite floats unchanged, re-raise the finite-overflow case) and it's already approved, so mine doesn't add anything. Closing in favor of #334. Thanks!

PR #344, opened 2026-07-04
naturaltime and precisedelta: return non-finite floats unchanged instead of raising OverflowError

HrachShah:
Both `naturaltime()` and `precisedelta()` feed a float input through `_date_and_delta()`, which passes it straight to `round()` (or skips rounding when `precise=True`). For `float('inf')` and `float('-inf')`, `round()` raises `OverflowError: cannot convert float infinity to integer` and `dt.timedelta(seconds=float('inf'))` raises the same. The except clause in the function catches `ValueError`, `TypeError`, but not `OverflowError`, so the user gets a traceback rather than the documented "return str(value) unchanged" behavior.

`naturaldelta()` already had the same bug, fixed in #342 by widening the except to include `OverflowError`. `naturaltime()` and `precisedelta()` still crash because they take a different path (`round()` instead of `int()`). This PR mirrors #342's fix for the two other public functions, by guarding the float-input branch with an explicit `math.isfinite` check before `round()`/`timedelta()` and by widening the except to `(ValueError, TypeError, OverflowError)` so a huge but finite value (e.g. `1e100`) falls through the same way.

`precise=True` (used by `precisedelta()`) skips the `round()` call but still constructs `dt.timedelta(seconds=value)`, which raises `OverflowError` for non-finite floats. The isfinite guard handles both paths.

Verified:
- `python3 -m pytest tests/` -> 724 passed, 69 skipped.
- 6 new test cases (3 each for naturaltime/precisedelta covering inf/-inf/nan) fail on the pre-fix code with `OverflowError: cannot convert float infinity to integer` and pass with the fix. 3 additional cases (the NaN parametrizations) pass on both pre-fix and post-fix because `round(float('nan'))` raises `ValueError`, which the original except already caught.
- Sanity check: `naturaltime(30)` still returns "30 seconds ago", `precisedelta(dt.timedelta(days=2, seconds=3633, microseconds=123000))` still returns "2 days, 1 hour and 33.12 seconds".

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -75,6 +75,7 @@ def _date_and_delta(
     If that's not possible, return `(None, value)`.
     """
     import datetime as dt
+    import math
 
     if not now:
         now = _now()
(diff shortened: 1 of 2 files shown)


hugovk:
There are now four open PRs to fix #333, and one is already an approved: https://github.com/python-humanize/humanize/pull/334.

hugovk:

PR #346, opened 2026-07-06
Fix `intword()` rounding carry for very large numbers

binggao1230:
### Problem

`intword()` decides whether rounding pushed a value up into the next magnitude
with:

```python
if not largest_ordinal and rounded_value * power == powers[ordinal + 1]:
```

For values above ~`10**22`, `rounded_value * power` is evaluated in floating
point and no longer equals the exact `powers[ordinal + 1]`, so the carry is
skipped and the number is rendered against the *lower* magnitude:

```python
>>> import humanize
>>> humanize.intword(10**24 - 1)
'1000.0 sextillion'      # expected '1.0 septillion'
>>> humanize.intword(10**27 - 1)
'1000.0 septillion'      # expected '1.0 octillion'
```

The same happens at `10**30` and `10**33`.

### Fix

Compare `rounded_value` against the exact integer ratio
`powers[ordinal + 1] // power`, keeping the check in exact integer terms rather
than relying on a large float product. This mirrors the recently fixed carry
handling in `metric()` (#328) and `naturalsize()` (#329).

### Tests

`test_intword_rounding_rollover` asserts the correct roll-over across every
affected magnitude (`10**24/27/30/33 - 1`), plus a couple of just-below cases
that must *not* roll over, guarding both directions.

`pytest` → 716 passed. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/src/humanize/number.py
+++ b/src/humanize/number.py
@@ -255,8 +255,12 @@ def intword(value: NumberOrString, format: str = "%.1f") -> str:
     chopped = value / power
     rounded_value = float(format % chopped)
 
-    if not largest_ordinal and rounded_value * power == powers[ordinal + 1]:
-        # After rounding, we end up just at the next power
+    if not largest_ordinal and rounded_value == powers[ordinal + 1] // power:
+        # After rounding, we end up just at the next power. Compare against the
+        # integer ratio between the two powers instead of ``rounded_value * power``:
+        # for values above ~10**22 the latter is evaluated in floating point and
+        # no longer equals the exact ``powers[ordinal + 1]``, so the carry was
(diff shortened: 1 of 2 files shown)


hugovk (approved):
```python
import humanize

print(humanize.intword(10**24 - 1))
print(humanize.intword(10**33 - 1))
print(humanize.intword(-(10**24 - 1)))
print(humanize.intword(10**24 - 1, "%.0f"))
```


# Before

```
1000.0 sextillion
1000.0 nonillion
-1000.0 sextillion
1000 sextillion
```

# After

```
1.0 septillion
1.0 decillion
-1.0 septillion
1 septillion
```

PR #349, opened 2026-07-09
filesize: raise ValueError for non-finite bytes in naturalsize

HrachShah:
**Bug**

`naturalsize(float('nan'))` raised `ValueError: cannot convert float NaN to integer` from inside the `int(min(log(...)))` exponent calculation, which is several levels removed from the call site. `naturalsize(float('inf'))` and `naturalsize(float('-inf'))` silently rendered as `'inf QB'` / `'-inf QB'`, which is not a meaningful byte count for any real caller.

**Fix**

Reject NaN, +inf, and -inf up front with a single `math.isfinite(bytes_)` check that names the rejected value in the error message. The same check covers string inputs that `float()` accepts but that don't represent a meaningful byte count (`'inf'`, `'INF'`, `'NaN'`, etc.); non-numeric strings still reach the existing `float()` call and raise ValueError as before.

**Tests**

Eight new cases in `tests/test_filesize.py`:
- `test_naturalsize_rejects_non_finite`: parametrized over `float('nan')`, `float('inf')`, `float('-inf')`.
- `test_naturalsize_rejects_non_finite_strings`: parametrized over `'nan'`, `'inf'`, `'-inf'`, `'NaN'`, `'INF'`.

Both assert that `pytest.raises(ValueError, match='finite number of bytes')` fires, so the fix is checked end-to-end on every relevant shape of non-finite input.

Three new doctest cases in `naturalsize`: `naturalsize(0)` (the `abs < base` path was previously untested), and explicit `Traceback` / `ValueError` cases for `naturalsize(float('nan'))` and `naturalsize(float('inf'))`.

Full suite: 723 passed, 69 skipped (no regressions).


hugovk:
Blocking and reporting @HrachShah for deleting CI here and https://github.com/python-attrs/attrs/pull/1585 and opening duplicate PRs (https://github.com/python-humanize/humanize/pull/342) and updating that PR while my question.

PR #351, opened 2026-07-14
Fix fractional(0) returning "0/1" instead of "0"

Sreekant13:
`fractional(0)` returns `"0/1"` instead of `"0"`, even though every other whole number renders bare:

```python
>>> import humanize
>>> humanize.fractional(0)
'0/1'   # expected '0'
>>> humanize.fractional(3)
'3'
```

The whole-number return branch was gated on `whole_number` being truthy, so `0` (where `whole_number == 0`) skipped it and fell through to the fraction branch, producing `"0/1"`. Since `numerator == 0 and denominator == 1` already uniquely identifies a whole number (including `0`), the `whole_number and` check is both redundant and wrong for `0`. Removing it makes `fractional(0)` return `"0"`, consistent with `fractional(3)`, `fractional(2.0)`, etc.

Fixes # _(no linked issue - small self-contained bug fix)_

* Drop the redundant `whole_number and` guard in `fractional()` so `0` is treated as a whole number and returns `"0"` (previously `"0/1"`).
* Add regression test cases `(0, "0")` and `(0.0, "0")` to `test_fractional`.


Sreekant13:
Hi @hugovk, gentle nudge on this small fix whenever you have time. Thanks for maintaining humanize.

Sreekant13:
Flagging an overlap, since I only noticed it now: @semx opened #354 a few days after this one, and it covers the same `fractional(0)` case plus a second one this PR does not touch, where the fractional part rounds up to a whole number and produces output like `"2 1/1"` for `fractional(2.9999999)`.

Their fix folds any whole-valued fractional part into the integer part, which handles both cases in one place, so it is the more complete change of the two. I am happy to close this in favour of #354 if that is the direction you prefer, or to add their case here if you would rather keep this one. Whichever is less work for you.

semx:
Thanks for the generous framing @Sreekant13 - for the record, your PR was here first, and priority counts for something too. Either direction works for me equally: if @hugovk prefers this one as first-come, I am happy to rebase #354 down to just the whole-rounding case on top of yours; if #354 goes in as the single change, credit for spotting fractional(0) first belongs here either way. Whatever is least work for the maintainer.

Sreekant13:
Good outcome. Thanks @hugovk for picking one, and thanks @semx for the generous note. #354 was the more complete fix, and it covers the case this PR found, so landing it is the right result.

PR #352, opened 2026-07-17
Fix ValueError in metric()/scientific() with precision 0

santhreal:
`metric(value, precision=0)` raises `ValueError: Format specifier missing precision` when the magnitude falls back to `scientific()`, because `metric` calls `scientific(value, precision - 1)`, so precision `0` becomes `-1` and builds an invalid format spec `"{:.-1e}"`. `scientific(value, -1)` directly hits the same path.

#159 fixed this for the non-scientific `metric` path; this covers the `scientific()` fallback it did not reach. `scientific` now clamps a negative precision to zero.

```python
import humanize
humanize.metric(1e300, "m", 0)   # was: ValueError; now: "1 x 10³⁰⁰m"
humanize.scientific(1e300, -1)   # was: ValueError; now: "1 x 10³⁰⁰"
```

### DIFF
--- a/src/humanize/number.py
+++ b/src/humanize/number.py
@@ -416,7 +416,8 @@ def scientific(value: NumberOrString, precision: int = 2) -> str:
             return _format_not_finite(value)
     except (ValueError, TypeError):
         return str(value)
-    fmt = f"{{:.{int(precision)}e}}"
+    # max(0): a negative precision builds an invalid format spec ("{:.-1e}").
+    fmt = f"{{:.{max(0, int(precision))}e}}"
     n = fmt.format(value)
     part1, part2 = n.split("e")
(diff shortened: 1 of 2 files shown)


hugovk:
You've opened many PRs across these repos two weeks ago, and then closed them with no comment:

https://github.com/python-humanize/humanize/pull/352
https://github.com/python-humanize/humanize/pull/353
https://github.com/prettytable/prettytable/pull/483
https://github.com/prettytable/prettytable/pull/484
https://github.com/jazzband/tablib/pull/659
https://github.com/jazzband/tablib/pull/660
https://github.com/jazzband/tablib/pull/661

Please could you explain why?

Also hundreds in other repos:

image

Issue #51, opened 2022-08-24
Filesize i18n not working for French


Benji81:
### What did you do?
```python
import humanize
humanize.i18n.activate("fr")
print(humanize.naturalsize(42))         # => '42 Bytes'
print(humanize.naturalsize(42_000))     # => '42.0 kB'
print(humanize.naturalsize(42_000_000)) # => '42.0 MB'
```

### What did you expect to happen?
```python
import humanize
humanize.i18n.activate("fr")
print(humanize.naturalsize(42))         # =>  '42 octets'
print(humanize.naturalsize(42_000))     # => '42,0 Ko'
print(humanize.naturalsize(42_000_000)) # => '42,0 Mo'
```
### What versions are you using?

* OS: ubuntu
* Python: 3.9.6
* Humanize:  4.3.0

carterbox:
This is not implemented. Other functions have translated output, but natural size does not. You're welcome to contribute a translation for this function.

https://github.com/python-humanize/humanize/blob/main/src/humanize/locale/fr_FR/LC_MESSAGES/humanize.po
https://github.com/python-humanize/humanize/blob/c6927228bed2fe90477537f96fa4ffc2564b5791/src/humanize/filesize.py

PR #357, opened 2026-07-23
Fix French decimal separator and apply to `naturalsize`

SpiliosDimakopoulos:
Re: #51


hugovk:
The decimal separator was set to `.` in https://github.com/python-humanize/humanize/pull/248, although https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma says France and Canada (when using French) use `,`

@merwok Shall we use `,` for French?

hugovk:
@SpiliosDimakopoulos Why did you close this?

merwok:
Definitely need `,` for fr_FR!

PR #358, opened 2026-07-25
fix: correct ordinal suffix for negative numbers (ordinal(-1) == "-1th")

MildlyMeticulous:
### The bug

```pycon
>>> humanize.ordinal(-1)
'-1th'
>>> humanize.ordinal(-2)
'-2th'
>>> humanize.ordinal(-3)
'-3th'
```

Expected `-1st`, `-2nd`, `-3rd`.

### Cause

```python
```

Python's `%` follows the sign of the divisor, not the dividend, so for negative input the digit index is the ten's complement of the actual last digit:

| value | `value % 10` | suffix picked | correct |
|---|---|---|---|
| -1 | 9 | th | **st** |
| -2 | 8 | th | **nd** |
| -3 | 7 | th | **rd** |
| -21 | 9 | th | **st** |
| -101 | 9 | th | **st** |

Note that `-11`, `-12`, `-13` come out right *by accident* — they map to 9, 8, 7, which are all `"th"` anyway — which is probably why this went unnoticed.

### Fix

```python
magnitude = abs(value)
digit = 0 if magnitude % 100 in (11, 12, 13) else magnitude % 10
```

The suffix follows the magnitude of the number, not its sign, so `abs()` is the whole fix. `-11`/`-12`/`-13` still land on the `(11, 12, 13)` teens branch and keep `"th"`, now for the right reason.

The `digit` index still ranges over 0-9, so `_ORDINAL_SUFFIXES` and the `P_(f'{digit} ({gender})', ...)` translation lookup are unaffected.

### Verification

10 negative cases added to the existing `test_ordinal` parametrization. Against the unpatched `number.py` **5 of them fail**; with the fix the full suite is **725 passed, 69 skipped, 0 failed**.

Behaviour for non-negative input, non-numeric input, `None`, `nan` and `±inf` is unchanged.


hugovk:
Duplicate of https://github.com/python-humanize/humanize/pull/321 and https://github.com/python-humanize/humanize/pull/339.

PR #360, opened 2026-07-25
Add Sinhala (si_LK) locale

RavinduPabasara:
Adds a Sinhala (`si_LK`) locale — the language of ~17 million speakers in Sri Lanka, previously unsupported by humanize.

* Add `src/humanize/locale/si_LK/LC_MESSAGES/humanize.po` translating all time, number-scale and file-size phrases into formal written Sinhala, verified by a native speaker.
* Add Sinhala to the localized-languages list in `README.md` (alphabetical position).
* Add Sinhala coverage to `tests/test_i18n.py` (`naturaltime`, `precisedelta`, `intword`, `naturalsize`).

Notes for reviewers:

* **Plural rule** follows CLDR for Sinhala: `nplurals=2; plural=(n > 1);`.
* **A grammatical subtlety worth knowing:** Sinhala marks "ago"/"from now" with a case suffix on the duration (`පැයකට පෙර` = "an hour ago", `පැය 2කට පෙර` = "2 hours ago"). Because gettext does plain substitution and cannot perform සන්ධි (morphophonemic joining), the single-unit forms (`an hour`, `a minute`, …) are rendered as bare nouns (`පැය`, `විනාඩිය`) so they compose correctly through `%s ago` / `%s from now`. This makes every `naturaltime` output grammatical; the only visible effect is that a direct `naturaldelta(timedelta(hours=1))` returns the bare noun `පැය` rather than an article form. Given `naturaltime` is the dominant consumer of these strings, this was the correct trade-off.

Full local run: `789 passed` (including the new Sinhala i18n assertions), `ruff`/`black` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ size or throughput. It is localized to:
 - Polish
 - Russian
 - Simplified Chinese
+- Sinhala
 - Slovak
 - Slovenian
 - Spanish
--- a/src/humanize/locale/si_LK/LC_MESSAGES/humanize.po
+++ b/src/humanize/locale/si_LK/LC_MESSAGES/humanize.po
@@ -1,0 +1,231 @@
+# Sinhala (si_LK) translations for the humanize package.
+# This file is distributed under the same license as the humanize package.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: humanize\n"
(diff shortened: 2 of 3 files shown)


hugovk:
Thank you!

PR #361, opened 2026-07-25
Let ordinal translations interpolate the number (#270)

### LINKED ISSUE #270: Impossible to translate to Romanian

RavinduPabasara:
Fixes #270

`ordinal()` builds its result as `f"{value}{suffix}"`, so a translation can only ever supply a *trailing* suffix. As reported in #270, that makes languages whose ordinals place the number *inside* the word impossible to translate — e.g. Romanian `al 2-lea` (m) / `a 2-a` (f), where the number sits between a prefix and a suffix.

* In `ordinal()`, if the (context-aware) translated string contains a `%s`/`%d` placeholder, format the number into it (`suffix % value`) so the translation controls where the number goes; otherwise keep the existing `"<number><suffix>"` behaviour.
* Add regression tests for both the interpolated path (Romanian-style `al 2-lea` / `a 2-a`, plus a `%d` variant) and the unchanged suffix path.

Backward compatibility:

* All 35 existing locales store bare suffixes (`"nd" -> "e"`, `"th" -> "º"`, …) with **no** placeholder, so they take the unchanged branch and their output is identical. The English default is unaffected (`ordinal(2) == "2nd"`). No `msgid` changes, so no existing translation needs re-merging.

Verified with a throwaway Romanian locale loaded through the real gettext path (`activate("ro", path=...)`):

```pycon
>>> humanize.ordinal(2)                  # al %s-lea
'al 2-lea'
>>> humanize.ordinal(23)                 # al %s-lea
'al 23-lea'
>>> humanize.ordinal(2, gender="female") # a %s-a
'a 2-a'
>>> humanize.i18n.deactivate()
>>> humanize.ordinal(2)
'2nd'
```

Full local run: `786 passed`, `ruff`/`black` clean.

### DIFF
--- a/src/humanize/number.py
+++ b/src/humanize/number.py
@@ -110,7 +110,14 @@ def ordinal(value: NumberOrString, gender: str = "male") -> str:
         return str(value)
     gender = "male" if gender == "male" else "female"
     digit = 0 if value % 100 in (11, 12, 13) else value % 10
-    return f"{value}{P_(f'{digit} ({gender})', _ORDINAL_SUFFIXES[digit])}"
+    suffix = P_(f"{digit} ({gender})", _ORDINAL_SUFFIXES[digit])
+    # Most languages append a suffix to the number ("2nd"), but some position the
+    # number inside the ordinal word (e.g. Romanian "al 2-lea", "a 2-a"). A
+    # translation can opt into that by including a "%s"/"%d" placeholder for the
+    # number; without one, the classic "<number><suffix>" form is kept.
+    if "%s" in suffix or "%d" in suffix:
+        return suffix % value
+    return f"{value}{suffix}"
 
 
 def intcomma(value: NumberOrString, ndigits: int | None = None) -> str:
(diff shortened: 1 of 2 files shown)


RavinduPabasara:
Closing this — I hadn't seen #331, which already addresses #270 (and ships the Romanian locale, including the special `primul`/`prima` forms). No need for a competing PR. Thanks!

PR #364, opened 2026-07-30
Fix `intword`'s decillion-to-googol gap and carry-over rounding

### LINKED ISSUE #356: intword: values between 10^36 and a googol format as huge decillion counts, and rounding to the next unit does not carry
### What happened

`intword` handles values up to 1000 decillion (10^36) and exactly a googol
(10^100) well, but everything in between formats as an enormous decillion
count:

```python
>>> import humanize
>>> humanize.intword(10**100)
'1.0 googol'
>>> humanize.intword(10**100 - 10**93)
'9999999000000000610151082626636586748864734715312950532670947852288.0 decillion'
>>> humanize.intword(10**50)
'100000000000000000.0 decillion'
```

A 67 digit unit count defeats the purpose of a human readable number, and
a value just below a googol that rounds up to it never carries to
"1.0 googol".

Verified on humanize 4.16.0 (published wheel) and on current main
(c3a124c), Python 3.11.

### Why this looks like a bug rather than a choice

This is the same class of boundary behavior that #328 recently fixed for
`metric()` (carry to the next SI prefix when rounding reaches 1000). The
powers list jumps from decillion (10^33) to googol (10^100), and the
selection loop picks decillion for anything under a googol, no matter how
large the resulting count is. Values above the largest power already fall
back to `str(value)`, which suggests the intended behavior for
"no reasonable unit" is the plain number, not a unit with a 67 digit
multiplier.

### Expected

One of the following, whichever matches the project's intent:

1. Values whose formatted count reaches 1000 of a unit with no adjacent
   next power fall back to `str(value)` (consistent with the existing
   above-maximum behavior), and values that round up to exactly the next
   power carry to it ("1.0 googol"), consistent with #328's carry rule.
2. Or simply fall back to `str(value)` for the whole gap between 10^36
   and 10^100.

I am happy to open a PR with regression tests for the boundary cases once
maintainers confirm which behavior is intended.

### How this was found

An automated review gate proposed edge case tests for the recent number.py
changes (the #328 carry fix among them) and executed them against the
code; the "carry just below the next unit" test failed with the output
above. The failing test and its observed output are reproducible by hand
as shown.

mmaxjr:
Fixes #356

## Problem
`intword()`'s `powers` table jumps straight from decillion (`10**33`) to googol (`10**100`), with no unit in between. Any value in [`10**36`, `10**100`) got chopped into a decillion count that grows arbitrarily large:

```pycon
>>> intword(10**50)
'100000000000000000.0 decillion'
>>> intword(10**100 - 10**93)
'9999999000000000610151082626636586748864734715312950532670947852288.0 decillion'


mmaxjr:
Status update on 2026-08-06: the latest commit (702e631) keeps the English fallback for unnamed magnitudes while preserving translated local-scale units such as French quintilliard(s).\n\nCurrent GitHub Actions runs for this fork update are in ction_required, so the full matrix needs maintainer approval to run. The external checks currently visible on the PR are green:\n\n- docs/readthedocs.org:humanize -> success\n- pre-commit.ci - pr -> success

hugovk:
Closing in favour of https://github.com/python-humanize/humanize/pull/346.

PR #367, opened 2026-07-31
Fix naturalsize() ValueError for custom formats containing text

### LINKED ISSUE #366: `naturalsize()` raises `ValueError` for custom format strings with text
## Environment

- humanize 4.16.0 and current `main` at `42b4a1db`
- Python 3.11.15

## Reproduction

```python
import humanize

print(humanize.naturalsize(999_999, gnu=True, format="Size: %.1f"))
```

humanize 4.15.0 returns:

```text
Size: 976.6K
```

humanize 4.16.0 and current `main` raise:

```text
ValueError: could not convert string to float: 'Size: 976.6'
```

## Expected behavior

The documented custom formatter should continue to accept text around the numeric conversion.
Since this behavior worked in 4.15.0, preserving it would avoid a backward compatibility regression.

## Cause

PR #329 added unit rollover handling. The new condition formats the mantissa with the
user-provided format and passes the entire result to `float()`:

```python
float(format % (abs_bytes / (base**exp)))
```

This works for a plain numeric format such as `"%.1f"`, but not for a general formatter
such as `"Size: %.1f"`.

## Suggested fix

It might be preferable not to require the complete user-formatted result to be parseable as a float. One
backward-compatible option is to apply rollover only when the formatted result is a
plain numeric string, and otherwise preserve the pre-4.16 behavior. A more complete
solution could derive the rounded numeric mantissa independently from surrounding text.

It might also be worth adding a regression test covering both a prefix and a suffix.

## How this was found

This was found by change-directed differential testing: reviewing the recent rollover change, mapping it to the public `naturalsize() API`, and constructing a documented custom formatter that reaches the newly added `float(format % value)` expression.

Related: #329, which fixes rollover but does not discuss custom format strings.

Thanks for taking the time to review this report! I appreciate all the work that goes into maintaining this project.

Sanjays2402:
Fixes #366

* `naturalsize()`'s unit-rollover check ran the user-supplied `format` through `float()`, so any format with text around the conversion (e.g. `"Size: %.1f"`) raised `ValueError` — a regression from 4.15.0.
* Compare the rendered mantissa against the rendered base instead, keeping the 999999 -> `1.0 MB` carry-over while accepting arbitrary formats again.
* Added two parametrized cases to `tests/test_filesize.py`; they fail with `ValueError` without the fix and pass with it.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/src/humanize/filesize.py
+++ b/src/humanize/filesize.py
@@ -103,7 +103,9 @@ def naturalsize(
     # mantissa afterward; rounding can push it up to `base` (e.g. 999999 is
     # 999.999 kB, which formats to "1000.0 kB"). When that happens and a larger
     # suffix is available, step up one suffix so the result reads "1.0 MB".
-    if exp < len(suffix) and abs(float(format % (abs_bytes / (base**exp)))) >= base:
+    # `format` may contain text around the conversion, so compare the rendered
+    # mantissa with the rendered base instead of parsing it back to a float.
+    if exp < len(suffix) and format % (abs_bytes / (base**exp)) == format % base:
         exp += 1
     space = "" if gnu else " "
     ret: str = format % (bytes_ / (base**exp)) + space + _(suffix[exp - 1])
--- a/src/humanize/filesize.py
+++ b/src/humanize/filesize.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 __lazy_modules__ = {"humanize.i18n", "math"}
 
+import re
 from math import log
 
 from humanize.i18n import _gettext as _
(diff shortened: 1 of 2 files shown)


MohammedAlkindi (commented):
The crash fix is right, but the new equality check changes output for `%e` formats in binary mode.

On this head, `naturalsize(1048575, True, False, "%.0e")` returns `'1e+00 MiB'`; at the merge base `42b4a1d` it returns `'1e+03 KiB'`. The unit now steps up at 972800 bytes instead of 1048576, so roughly 7% of the KiB range renders one unit high. Negative sizes do the same.

Cause: `format % x == format % base` assumes `float(format % base) == base`. For `%.0e` with base 1024 it does not, since `"%.0e" % 1024` is `'1e+03'`. Decimal mode is unaffected because 1000 round-trips exactly.

Suite is clean either way: 715 at the merge base, 717 here, the difference being your two added cases.

Sanjays2402:
Thanks @MohammedAlkindi — you were right, the string comparison was wrong for formats that don't round-trip through str -> float. Reworked the check: instead of comparing `format % x == format % base`, it now extracts the numeric mantissa from the rendered string (so formats with surrounding text still don't crash on `float()`) and compares that numerically against `base`. `naturalsize(1048575, True, False, "%.0e")` is back to `'1e+03 KiB'`, step-up stays at 1048576, and the original text-format crash cases still pass. Added a regression test for the `%.0e` case.

PR #373, opened 2026-08-02
fix(time): catch OverflowError for non-finite floats (#333)

### LINKED ISSUE #333: naturaldelta() raises OverflowError on float('inf') instead of returning it unchanged

MeiSiristhebest:
## Description
Fixes #333 where `naturaldelta()`, `naturaltime()`, and `precisedelta()` raise an uncaught `OverflowError` when passed `float('inf')` or `float('-inf')`.

## Proposed Solution
Expand exception handling in `_date_and_delta()` and `naturaldelta()` from `(ValueError, TypeError)` to `(ValueError, TypeError, OverflowError)`.

## Testing Evidence
- Added tests in `tests/test_time.py` covering `float('inf')` and `float('-inf')`.
- Full test suite passed (719 passed, 74 skipped, 0 failed).

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 __lazy_modules__ = {"humanize.i18n", "humanize.number"}
 
+import math
 from enum import Enum
 from functools import total_ordering
 
@@ -89,8 +90,12 @@ def _date_and_delta(
             value = value if precise else round(value)
             delta = dt.timedelta(seconds=value)
             date = now - delta
-        except (ValueError, TypeError, OverflowError):
+        except (ValueError, TypeError):
             return None, value
+        except OverflowError:
+            if not math.isfinite(value):
+                return None, value
+            raise
     return date, _abs_timedelta(delta)
 
 
@@ -116,6 +121,9 @@ def naturaldelta(
             elapsed unless `value` is not datetime.timedelta or cannot be
             converted to int (cannot be float due to 'inf' or 'nan').
             In that case, a `value` is returned unchanged.
+
+    Raises:
+        OverflowError: If `value` is too large to convert to datetime.timedelta.
 
     Examples:
         Compare two timestamps in a custom local timezone::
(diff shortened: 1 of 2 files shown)


hugovk:
Why is this better than the other 5 or so PRs opened for #333?

MeiSiristhebest:
Good point. #334 handles `naturaldelta`, but `naturaltime` and `precisedelta` also crash on `inf` because they go through `_date_and_delta`:

```pycon
>>> humanize.naturaltime(float("inf"))
>>> humanize.precisedelta(float("inf"))
```

This PR patches `_date_and_delta` so all three functions handle non-finite floats.

You're right that catching `OverflowError` broadly was wrong — it suppressed exceptions for large finite floats like `1e30`. I've updated this PR to match #334's check using `math.isfinite()`, restored the docstring, and added a test for the finite overflow behavior.

If you'd prefer to fold the `_date_and_delta` fix into #334 instead, I'm also happy to close this one.

Mukller:
Heads-up for maintainers: this PR and #375 fix the same OverflowError-on-non-finite issue (#333) but with **opposite semantics for finite too-large values**:

- **This PR (#373)**: returns the value unchanged only for non-finite inputs (`inf`/`-inf`); a *finite* too-large value still raises the documented `OverflowError`. The added test `test_naturaldelta_too_large_value_raises` pins that contract.
- **#375**: adds `OverflowError` to the existing `except (ValueError, TypeError)` catch-alls in `naturaldelta`/`naturaltime`, so *any* oversized value silently becomes `str(value)` — removing the documented "Raises: OverflowError" contract entirely.

These will behave differently for e.g. `naturaldelta(1e20)` (finite, exceeds timedelta range): this PR raises, #375 returns `"1e+20"`. Whichever semantics the maintainer prefers, one of the two should be closed as duplicate — flagging early since neither references the other.

(Also noting for completeness that I reviewed #374 separately; unrelated.)

PR #375, opened 2026-08-09
fix: catch OverflowError for non-finite floats in time functions

### LINKED ISSUE #333: naturaldelta() raises OverflowError on float('inf') instead of returning it unchanged

Manushprajwal7:
## Summary

Fixes #333.

`naturaldelta()`, `naturaltime()`, and `precisedelta()` all raise an uncaught `OverflowError` for `float('inf')` / `float('-inf')` instead of returning the value unchanged — inconsistent with every other numeric humanize function (`ordinal`, `intcomma`, `intword`, etc.), which already treat non-finite input this way, and inconsistent with how these same three functions already handle `float('nan')`.

```pycon
>>> import humanize
    ...
```

**Root cause:** `int()` / `round()` raise `OverflowError` (not `ValueError` or `TypeError`) for infinite floats. That exception wasn't included in the `except` clauses guarding the `timedelta` conversion in `naturaldelta()`, nor in the shared `_date_and_delta()` helper used by `naturaltime()` and `precisedelta()`. Since both call sites share the same root cause, this fixes all three functions rather than just the one named in the issue.

## Changes

- `src/humanize/time.py`: add `OverflowError` to the two relevant `except` clauses (in `naturaldelta()` and `_date_and_delta()`).
- Updated `naturaldelta()`'s docstring, which documented the crash as a `Raises: OverflowError` — that was describing the bug, not an intentional design.
- Added regression tests for `inf`/`-inf`/`nan` across `naturaldelta`, `naturaltime`, and `precisedelta`.

## Test plan

- [x] `pytest tests/test_time.py -q` — 392 passed
- [x] Full suite: `pytest -q` — 724 passed, 74 skipped, 0 failed
- [x] `ruff check` on changed files — clean
- [x] Manually verified `naturaldelta`/`naturaltime`/`precisedelta` no longer raise for `inf`/`-inf`, and behavior for genuinely-too-large-but-finite values (e.g. `1e300`) is now consistent (returned unchanged rather than raising)

### DIFF
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -146,7 +146,9 @@ def naturaldelta(
         delta = value
     else:
         try:
-            int(value)  # Force numeric types; strings like "NaN"/"inf" are returned unchanged
+            int(
+                value
+            )  # Force numeric types; strings like "NaN"/"inf" are returned unchanged
             value = float(value)
             delta = dt.timedelta(seconds=value)
         except (ValueError, TypeError, OverflowError):
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -146,7 +146,7 @@ def naturaldelta(
         delta = value
     else:
         try:
-            int(value)  # Explicitly don't support string such as "NaN" or "inf"
+            int(value)  # Force numeric types; strings like "NaN"/"inf" are returned unchanged
             value = float(value)
             delta = dt.timedelta(seconds=value)
         except (ValueError, TypeError, OverflowError):
--- a/src/humanize/time.py
+++ b/src/humanize/time.py
@@ -89,7 +89,7 @@ def _date_and_delta(
             value = value if precise else round(value)
             delta = dt.timedelta(seconds=value)
             date = now - delta
-        except (ValueError, TypeError):
+        except (ValueError, TypeError, OverflowError):
             return None, value
     return date, _abs_timedelta(delta)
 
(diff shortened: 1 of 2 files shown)


Mukller:
Heads-up for maintainers: this PR and #375 fix the same OverflowError-on-non-finite issue (#333) but with **opposite semantics for finite too-large values**:

- **#373**: returns the value unchanged only for non-finite inputs (`inf`/`-inf`); a *finite* too-large value still raises the documented `OverflowError`, with a test pinning that contract.
- **This PR (#375)**: adds `OverflowError` to the existing `except (ValueError, TypeError)` catch-alls in `naturaldelta`/`naturaltime`, so *any* oversized value silently becomes `str(value)` — removing the documented "Raises: OverflowError" contract entirely.

These will behave differently for e.g. `naturaldelta(1e20)` (finite, exceeds timedelta range): this PR raises, #375 returns `"1e+20"`. Whichever semantics the maintainer prefers, this PR and #373 diverge as described above, so one should likely be closed or reworked once semantics are chosen — flagging early since neither references the other.

(Also noting for completeness that I reviewed #374 separately; unrelated.)

hugovk:
Something like the 7th duplicate for #333. Please check next time and don't waste time with duplicates.

PR #376, opened 2026-08-11
fix(number, lists): correct negative ordinal suffixes and support general iterables with oxford_comma

miskatul-anwar:
## Summary

This PR resolves two key issues and enhances developer ergonomics:

1. **Negative Integer Ordinal Suffix Calculation (`src/humanize/number.py`)**:
   - Fixed remainder calculations in `ordinal()` for negative integers (e.g. `-21` previously returned `'-21th'` due to Python's modulo behavior on negative numbers). Using `abs(value)` ensures correct suffixes (e.g., `-1st`, `-2nd`, `-3rd`, `-11th`, `-21st`).

2. **Iterable & Oxford Comma Support in `natural_list` (`src/humanize/lists.py`)**:
   - Updated `natural_list` to accept any general `Iterable` (such as generators, tuples, sets) instead of strictly requiring `list`.
   - Added an optional `oxford_comma: bool = False` parameter to format lists as `'one, two, and three'`.

3. **Source Import & Test Resilience (`src/humanize/__init__.py`, `tests/`)**:
   - Added a `try...except ImportError` fallback for `_version.py` in `__init__.py` so importing directly from source without build artifacts works smoothly.
   - Added `pytest.importorskip` for optional test dependencies (`freezegun`, `pytest-codspeed`) to prevent test crashes in minimal development environments.

## Empirical Test Verification
Ran `pytest`:
```text
319 passed, 3 skipped in 0.39s
```


Mukller (commented):
Verified locally on the PR branch vs release 4.16.0 (Python 3.13):

**The negative-ordinal fix is real and correct** — Python's modulo makes the old code produce wrong suffixes:

| input | release | branch |
|---|---|---|
| `ordinal(-1)` | `-1th` | `-1st` |
| `ordinal(-2)` | `-2th` | `-2nd` |
| `ordinal(-11)` / `(-111)` | `-11th` / `-111th` | same (11/12/13 rule preserved via `abs % 100`) |

`natural_list` improvements verified too: plain generators now work (release raises `TypeError`), `oxford_comma=True` yields `"a, b, and three"`-style output while default stays byte-identical, two-item lists ignore the flag correctly.

`pytest tests/test_lists.py tests/test_number.py` — **243 passed**.

One request before merge, on an unrelated bundled hunk: the `__version__` fallback hardcodes `"4.12.2.dev0"` when `_version.py` is missing. That misreports the actual version in exactly the broken environments where accurate versions matter most (bug reports from source checkouts would claim 4.12.2). Suggest either dropping that hunk from this PR (it's orthogonal to both fixes here) or using a neutral placeholder via `importlib.metadata.version("humanize")` with a fallback of `"unknown"`.

Everything else: approve-worthy.

miskatul-anwar:
Thanks for the review @Mukller! I have dropped the `_version.py` fallback hunk from this PR as suggested.

hugovk:
Duplicate of #321 and #381, and adds a new feature that doesn't belong in a fix. Please check for duplicates next time to avoid wasting our time.

PR #382, opened 2026-08-26
Remove the leftover format placeholder from `intword` units

VenishPaneliya:
`intword` inserts the unit without any `%` substitution:

```python
singular, plural = human_powers[ordinal]
unit = _ngettext(singular, plural, math.ceil(rounded_value))
...
return f"{negative_prefix}{number} {unit}"
```

so a printf placeholder left in a translated unit is printed literally. The Bengali, Korean and Vietnamese catalogues still carry `%d` or `%(value)s` on the `thousand`…`googol` units:

```python
humanize.i18n.activate("bn_BD"); humanize.intword(1_500_000_000)
humanize.i18n.activate("vi_VN"); humanize.intword(1_200_000)
humanize.i18n.activate("ko_KR"); humanize.intword(1_200_000)
```

On `main`:

```
'1.5 %d বিলিয়ন'
'1.2 %(value)s triệu'
'1.2 %(value)s million'
```

With this PR:

```
'1.5 বিলিয়ন'
'1.2 triệu'
'1.2 million'
```

This strips the placeholder and the separator that follows it (a plain space or U+00A0) and leaves the translated words untouched.

Scope is deliberately narrow: only the twelve `intword` power msgids are changed. Placeholders elsewhere in these catalogues are substituted normally and are left alone — `naturaldelta`'s `"%d day"` and friends still need theirs. After the change, none of the 36 catalogues emits a placeholder for any of the twelve powers.

Added a parametrised regression test over the three locales, checking that `intword` never returns a `%` at any power. It fails 3/3 on `main` and passes here.

Two notes I did not act on, since both need a native speaker rather than a scripted edit:

- `ko_KR` renders several powers as the English words (`million`, `billion`), and its `trillion` entry reads `billion`.
- `ko_KR` and `vi_VN` also contain a couple of French leftovers (`décillion`, `gogol`).

Happy to split those out separately if you would like them addressed.

### DIFF
--- a/tests/test_i18n.py
+++ b/tests/test_i18n.py
@@ -334,3 +334,14 @@ class TestActivate:
             humanize.i18n.deactivate()
 
         assert test_str == humanize.naturaltime(three_seconds)
+
+
+@pytest.mark.parametrize("locale", ["bn_BD", "ko_KR", "vi_VN"])
+def test_intword_unit_has_no_format_placeholder(locale: str) -> None:
+    """`intword` interpolates the unit verbatim, so it must carry no placeholder."""
+    try:
+        humanize.i18n.activate(locale)
+        for value in (1_000, 1_000_000, 1_000_000_000, 10**12, 10**15, 10**100):
+            assert "%" not in humanize.intword(value)
+    finally:
+        humanize.i18n.deactivate()
--- a/src/humanize/locale/bn_BD/LC_MESSAGES/humanize.po
+++ b/src/humanize/locale/bn_BD/LC_MESSAGES/humanize.po
@@ -121,57 +121,57 @@ msgstr "তম"
 #: src/humanize/number.py:178
 msgid "thousand"
 msgid_plural "thousand"
-msgstr[0] "%d হাজার"
-msgstr[1] "%d হাজার"
+msgstr[0] "হাজার"
+msgstr[1] "হাজার"
 
 #: src/humanize/number.py:179
 #, fuzzy
 msgid "million"
 msgid_plural "million"
-msgstr[0] "%d মিলিয়ন"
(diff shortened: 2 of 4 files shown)


tests/test_i18n.py
```diff
@@ -334,3 +334,14 @@ def test_none_locale(self) -> None:
...
+
+@pytest.mark.parametrize("locale", ["bn_BD", "ko_KR", "vi_VN"])
+def test_intword_unit_has_no_format_placeholder(locale: str) -> None:
+    """`intword` interpolates the unit verbatim, so it must carry no placeholder."""
+    try:
+        humanize.i18n.activate(locale)
+        for value in (1_000, 1_000_000, 1_000_000_000, 10**12, 10**15, 10**100):
+            assert "%" not in humanize.intword(value)
+    finally:
+        humanize.i18n.deactivate()
```
hugovk: Let's test all locales, and also add a skip when file not found (for example, Windows on CI):

```suggestion
@pytest.mark.parametrize(
    "locale", sorted(p.name for p in LOCALE_DIR.iterdir() if p.is_dir())
)
def test_intword_unit_has_no_format_placeholder(locale: str) -> None:
    try:
        humanize.i18n.activate(locale)
        for value in (1_000, 1_000_000, 1_000_000_000, 10 ** 12, 10 ** 15, 10 ** 100):
            assert "%" not in humanize.intword(value)
    except FileNotFoundError:
        pytest.skip("Generate .mo with scripts/generate-translation-binaries.sh")
    finally:
        humanize.i18n.deactivate()
```

hugovk (approved):

hugovk:
Thanks!

PR #383, opened 2026-08-29
Profile contributors

hugovk:
Help cope with all the LLM contributions coming in.

These are based on the output of `uvx gh-profiler --generate-workflow`, but I don't want any `pull_request_target` here, so:

* `profile-issue-authors.yml` is a stripped down version of the generated one, but only for issues - comments on issues, and deletes comments later
* `profile-pr-authors.yml` cannot write or delete comments on PRs, so it's shown in the PR actions summary instead, and expires with the logs


hugovk:
Example: https://github.com/python-humanize/humanize/actions/runs/33263070382?pr=383

PR #390, opened 2026-09-05
Handle non-finite values in naturalsize

AnishPatel526:

* `naturalsize()` now handles non-finite input consistently with the rest of the library. Every numeric formatter in `humanize.number` (`ordinal`, `intcomma`, `intword`, `apnumber`, `fractional`, `scientific`, `clamp`) already returns `"NaN"` / `"+Inf"` / `"-Inf"` for non-finite values via `_format_not_finite`. `naturalsize()` was the exception: `naturalsize(float("nan"))` raised `ValueError: cannot convert float NaN to integer`, and `naturalsize(float("inf"))` produced the bogus string `"inf QB"`. It now returns `"NaN"` / `"+Inf"` / `"-Inf"`.
* Added a parametrized `test_naturalsize_non_finite` covering `nan`, `inf`, `-inf` as both floats and strings; it fails on `main` and passes with this change.

### Before

```pycon
>>> import humanize
>>> humanize.naturalsize(float("nan"))
ValueError: cannot convert float NaN to integer
>>> humanize.naturalsize(float("inf"))
'inf QB'
```

### After

```pycon
>>> humanize.naturalsize(float("nan"))
'NaN'
>>> humanize.naturalsize(float("inf"))
'+Inf'
>>> humanize.naturalsize(float("-inf"))
'-Inf'
```

The fix reuses the existing `humanize.number._format_not_finite` helper so output matches the other formatters exactly, and guards right after the `float(value)` conversion. Finite values are unaffected.

### DIFF
--- a/src/humanize/filesize.py
+++ b/src/humanize/filesize.py
@@ -4,9 +4,10 @@ from __future__ import annotations
 
 __lazy_modules__ = {"humanize.i18n", "math"}
 
-from math import log
+from math import isfinite, log
 
 from humanize.i18n import _gettext as _
+from humanize.number import _format_not_finite
 
 suffixes = {
     "decimal": (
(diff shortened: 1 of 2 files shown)


hugovk:
Closing as duplicate of of #299.

Check for duplicates yourself next time.

PR #393, opened 2026-09-07
Hoist `import math` to module scope in number.py

g4lb:
### What

#335 added `__lazy_modules__` so that module-scope imports become lazy on Python
3.15+. It added the declarations only — the function-level `import math` statements
that predated it were left in place. This finishes that migration for `number.py`:
hoists the nine function-level `import math` statements to module scope, and adds
`"math"` to `__lazy_modules__`, matching what `filesize.py` already does for the very
same module.

### Why

A function-level `import` still costs a `sys.modules` lookup and a name binding on
**every call**. It buys nothing here, because `math` is already resident by the time
`humanize` finishes importing — `python -X importtime -c "import humanize"` shows
`math` loading at ~425 µs, before `humanize.i18n`.

Measured with `pytest-benchmark` on the repository's own `tests/test_benchmarks.py`
(Apple Silicon, CPython 3.14.6, medians of interleaved runs):

| benchmark | before | after | change |
|---|---:|---:|---:|
| `test_clamp` | 177.1 ns | 134.9 ns | **−23.8 %** |
| `test_apnumber` | 304.7 ns | 252.0 ns | **−17.3 %** |
| `test_ordinal` | 417.7 ns | 368.7 ns | −11.7 % |
| `test_scientific` | 458.0 ns | 417.0 ns | −9.0 % |
| `test_intword` | 958.0 ns | 875.0 ns | −8.7 % |
| `test_metric` | 500.0 ns | 459.0 ns | −8.2 % |

Geometric mean across all 15 benchmarks: **0.9409 (−5.9 %)**, no benchmark regressed.

### Import time

Unchanged, as expected given `math` is already loaded:

- before: 1.64 / 1.55 / 1.58 ms
- after: 1.56 / 1.51 / 1.58 ms

(`import humanize` in a fresh interpreter, best of five, three repetitions.)

### Checks

`pytest tests/` → 724 passed, 74 skipped. `ruff check`, `black --check` and `mypy`
all clean on the changed file.

### Not included

`time.py` has the same leftover pattern with `import datetime as dt` in eight
functions. It is a real win on the time benchmarks (`test_naturaltime` −13.6 %,
`test_naturalday` −6.6 %, `test_naturaldate` −6.3 %) but only −0.75 % on the geomean
across all 15, so I have left it out to keep this change to one idea. Happy to open it
separately if you would like it.

### How this was found

By pointing [autor3search-python](https://github.com/g4lb/autor3search-python) at
`humanize` — a harness that proposes a change, measures it against a pinned baseline
with interleaved A/B rounds, and keeps or discards it on the numbers. `humanize` was
the first real library it has been run against. Every figure above is a measurement,
not an estimate.


hugovk:
> A function-level `import` still costs a `sys.modules` lookup and a name binding on
> **every call**. It buys nothing here, because `math` is already resident by the time
> `humanize` finishes importing — `python -X importtime -c "import humanize"` shows
> `math` loading at ~425 µs, before `humanize.i18n`.

I see the deferred `math` import had no effect because we were still importing it eagerly elsewhere. 

https://github.com/python-humanize/humanize/pull/397 defers it there too, and means we also get the benefit for 3.14 and older.

We don't get the improvements from this PR, avoiding the `import math` cache lookup, but these are all small numbers so it's okay.

PR #395, opened 2026-09-07
fix precisedelta dropping sign of negative timedeltas (fixes #379)

### LINKED ISSUE #379: precisedelta() silently drops the sign of negative timedeltas
`precisedelta()` **silently drops the sign** of negative timedeltas — the output is byte-identical to the positive counterpart, so a negative delta rendered in any UI becomes indistinguishable from a positive one.

**Version:** humanize 4.16.0 / current master. Python 3.13.

## Repro

```python
import datetime as dt
from humanize import precisedelta

td = dt.timedelta(seconds=-3661)

print(precisedelta(td))                    # '1 hour, 1 minute and 1 second'
print(precisedelta(dt.timedelta(seconds=3661)))  # '1 hour, 1 minute and 1 second'  <- identical!
```

The sign is lost for every unit combination (also checked `minimum_unit="minutes"`). Neither the docstring nor the tests mention negative inputs at all, so it's unclear whether this is intended — but the current behavior makes the function unsafe for anything that can produce signed deltas (countdown corrections, progress deltas, A/B time differences), because the rendered string carries no trace of the direction.

For comparison, sibling function `naturaldelta()` also uses absolute value internally, but its output vocabulary ("a moment", "an hour") reads as inherently directionless; `precisedelta`'s numeric output does not.

## Suggested resolutions (either would work)

1. **Preserve the sign**: prefix the result with `-` when `value` is negative (mirroring `str(timedelta)`, which keeps the sign: `str(td)` → `'-1 day, 1:01:01'`).
2. Or raise / document explicitly that only non-negative values are supported.

Option 1 seems most useful and backwards-compatible for positive inputs; option 2 at least removes the silent-corruption trap.

Happy to send a PR with either behavior + tests if maintainers indicate a preference.

aDragon0707:
Fixes #379.

`precisedelta()` silently dropped the sign of negative timedeltas — the output was byte-identical to the positive counterpart (`1 hour, 1 minute and 1 second` for both `-3661` and `+3661` seconds).

- Detect a negative value before `_date_and_delta`, then prefix `-` on the rendered result.
- Zero stays `0 seconds` (no `-0`).
- Positive inputs are unchanged.

Regression test added (`test_precisedelta_negative`): seconds, positive control, and `minimum_unit="minutes"`.

`tests/test_time.py` → 388 passed, 0 failed.


hugovk:
> Fixes #379.

If you open this issue, you'll see there's already an open PR for this: https://github.com/python-humanize/humanize/pull/380.

Please check next time, I'm getting so many duplicates that it wastes review time that can be spent elsewhere.

PR #396, opened 2026-09-07
fix naturalsize ValueError for custom format strings with text (fixes #366)

### LINKED ISSUE #366: `naturalsize()` raises `ValueError` for custom format strings with text

aDragon0707:
Fixes #366.

`naturalsize()` raised `ValueError` for custom format strings with surrounding text (e.g. `format="Size: %.1f"`), because the rollover check passed the full formatted result to `float()`.

- The rollover step now tries to parse the formatted mantissa as a float and skips the rollover when the formatted result is not a plain number, preserving the pre-4.16 behavior for text formats.
- Plain numeric formats keep the rollover behavior (`999_999 -> "1.0 MB"` still works).

Regression test added (`test_naturalsize_custom_format_with_text`): prefix and suffix cases.

`tests/test_filesize.py` → 77 passed, 0 failed.


tests/test_filesize.py
```diff
@@ -103,3 +103,14 @@ def test_naturalsize(test_args: list[int] | list[int | bool], expected: str) ->
...
+
+
+def test_naturalsize_custom_format_with_text() -> None:
+    # regression: custom format strings with surrounding text must not crash
+    # https://github.com/python-humanize/humanize/issues/366
+    assert (
+        humanize.naturalsize(999_999, gnu=True, format="Size: %.1f") == "Size: 976.6K"
+    )
+    assert (
+        humanize.naturalsize(999_999, gnu=True, format="%.1f bytes") == "976.6 bytesK"
```
hugovk, line 115: "976.6 bytesK" is nonsense! 🙃

hugovk:
> Fixes #366.

If you open this issue, you'll see there's already an open PR for this: https://github.com/python-humanize/humanize/pull/367.

Check next time, I'm getting so many duplicates that it wastes review time that can be spent elsewhere.

PR #397, opened 2026-09-08
Defer importing `math` in `filesize.py`

hugovk:
Alternative to and closes #393.

We were already deferring import of `math` in `number.py` (https://github.com/python-humanize/humanize/pull/238), but this was negating that.

### DIFF
--- a/src/humanize/filesize.py
+++ b/src/humanize/filesize.py
@@ -2,9 +2,7 @@
 
 from __future__ import annotations
 
-__lazy_modules__ = {"humanize.i18n", "math"}
-
-from math import log
+__lazy_modules__ = {"humanize.i18n"}
 
 from humanize.i18n import _gettext as _
 
@@ -97,6 +95,8 @@ def naturalsize(
 
     if abs_bytes < base:
         return f"{int(bytes_)}B" if gnu else _("%d Bytes") % int(bytes_)
+
+    from math import log
 
     exp = int(min(log(abs_bytes, base), len(suffix)))
     # The suffix is chosen from the unrounded byte count, but `format` rounds the


AbhiramMandala:
## What changed

- `src/humanize/filesize.py` only: `__lazy_modules__` reduced from `{"humanize.i18n", "math"}` to `{"humanize.i18n"}`, and `from math import log` moved from module scope into the `naturalsize()` body (the `abs_bytes >= base` branch).
- Net diff is +3/-3 in a single file.

## Why this approach

- `__lazy_modules__` defers the load only on Python 3.15+. On the supported 3.10-3.14 runtimes a module-scope `from math import log` eagerly imports `math` at `import humanize` time, which negates the deferral `number.py` already uses (function-level `import math`, PR #238).
- A function-level import defers the module load on **all** supported Pythons without relying on the 3.15-only lazy-import machinery, and matches the existing `number.py` convention.

## Validation

- `pytest` (full `tests/`, incl. `test_filesize`) ? **724 passed, 74 skipped** on both Python 3.14.7 and Python 3.15.0rc2 (PR head checked out).
- `"math" in sys.modules` immediately after `import humanize`: **main = True** (eagerly imported), **PR head = False** (deferred).
- After `humanize.naturalsize(...)`: `math` present in `sys.modules` ? deferral holds exactly until first call.
- `naturalsize` outputs spot-checked: `3.0 MB`, `300B` (gnu), `2.9 KiB` (binary), `-4.0 KiB`, `1 Byte` ? unchanged.
- CI on the PR is fully green (test, lint, mypy, build, codspeed, codecov).

## Scope

- Only `src/humanize/filesize.py` touched; no behavioral, import-order, or dependency changes.

## Reviewer focus

- The import is now inside `naturalsize()`, in the branch only reached when `abs_bytes >= base`; confirm codspeed/benchmarks show no measurable per-call regression (function-level import is a `sys.modules` lookup + local binding).

PR #399, opened 2026-09-15
Fix ordinal suffix for negative integers

Aryan-Kochhar:

* Fix `ordinal()` returning the wrong suffix for negative integers. The suffix came from `value % 10`, and Python's modulo is non-negative for a positive divisor, so `-1 % 10` is `9` and the lookup landed on `"th"`. Every negative ending in 1, 2 or 3 was affected.
* Take the suffix from `abs(value)` instead. Positives are untouched, and the teens rule still holds because `abs(-11) % 100` is still `11`.
* Add negative integers to the parametrised test, which previously only had `-math.inf` and `"-inf"`.

Before:

```pycon
>>> humanize.ordinal(-1)
'-1th'
>>> humanize.ordinal(-2)
'-2th'
>>> humanize.ordinal(-21)
'-21th'
```

After:

```pycon
>>> humanize.ordinal(-1)
'-1st'
>>> humanize.ordinal(-2)
'-2nd'
>>> humanize.ordinal(-21)
'-21st'
>>> humanize.ordinal(-11)
'-11th'
```

The docstring says the function works for any integer, and negatives do reach the suffix lookup
rather than falling back to `str(value)`, so they came out wrong rather than being rejected.

Five of the new cases (`-1`, `-2`, `-3`, `-21`, `-101`) fail without the change; `-4`, `-11`, `-12`,
`-13` and `-111` are guards that the `"th"` rule still holds.


hugovk:
Please check for duplicates (https://github.com/python-humanize/humanize/pull/321) before opening new PRs. This is now the fourth such duplicate.

Aryan-Kochhar:
Apologies, I should have searched before opening this. Closing in favour of #321.

hugovk:
But I already closed it.

PR #404, opened 2026-09-18
Preserve large integer strings in intcomma

x0Lazarus:
`intcomma(str(10**400 + 123))` currently returns `"+Inf"`, although the same finite value passed as an integer is formatted correctly. The string path checks `float(value)` before attempting integer parsing, so values beyond the floating-point range are mistaken for infinity.

With the default precision, parse integer strings with `int()` first, after normalizing the locale's separators. Inputs that are not integer strings continue through the existing float/non-finite handling. This extends the default-precision behavior fixed for integer inputs in #392 to plain and already-grouped integer strings.

The regression tests cover both signs, preserve the final `123`, and check German and French grouping. Compatibility cases retain the existing handling of decimal strings, exponent strings, invalid input, and non-finite values. The existing float check is retained when `ndigits` is explicitly set, with controls for both signs beyond the float range.

Validation on Windows with Python 3.12:

- Four plain/grouped string regressions fail on the original source and pass with the change.
- Focused number and localization tests: 49 passed.
- Full suite with doctests and benchmark timing disabled: 799 passed, 69 skipped. German, French and Portuguese catalogs were compiled; remaining translation cases and gettext tool checks were skipped.
- Ruff and Black pass for the three changed files; mypy passes for all 14 source/test files; `git diff --check` passes.


Voyagerroc-Lab (approved):
Reviewed 6cc366f. Parsing integer strings before converting them to float fixes the overflow-to-infinity case while retaining existing decimal, exponent, and non-finite handling. Added cases cover both signs, grouped strings, and German/French separators. GitHub checks for the current commit pass. The PR author reports 49 focused tests and 799 passed / 69 skipped in the full suite, plus Ruff, Black, mypy, and diff checks; I did not rerun the suite locally. Review performed with OpenAI Codex at the account owners request.

hugovk:
@Voyagerroc-Lab There's no need to use AI to summarise the original PR description.

PR #405, opened 2026-09-18
Fix ordinal() suffix for negative numbers

itzzdev09:
Calling `ordinal()` on a negative integer gives the wrong suffix:

```pycon
>>> ordinal(-1)
'-1th'
>>> ordinal(-9)
'-9st'
>>> ordinal(-21)
'-21th'
```

The suffix is chosen from `value % 10` and `value % 100`, but Python's modulo of a negative number counts up from the next lower multiple (`-9 % 10 == 1`, `-1 % 10 == 9`), so the digit used for the lookup is wrong for every negative value.

Computing the digit from `abs(value)` fixes it while leaving non-negative values, the 11/12/13 special case, and the non-finite / non-numeric paths unchanged:

```pycon
>>> ordinal(-1)
'-1st'
>>> ordinal(-9)
'-9th'
>>> ordinal(-21)
'-21st'
```

Added the negative cases to `test_ordinal`; the full `test_number.py` suite passes.


hugovk:
Please see questions on duplicate #321.

itzzdev09:
I think the framing on #321 was "do negative ordinals make sense," but that's not quite what this fixes. ordinal() already accepts negative ints today — it doesn't raise — it just returns grammatically impossible strings: ordinal(-9) gives -9st, ordinal(-21) gives -21th. So the choice isn't "support negatives or not," it's "keep emitting broken output, or fix it." If you'd genuinely rather not handle negatives, I'd argue raising or returning the value unchanged is cleaner than returning -9st. But given it already accepts them, correcting the suffix seemed like the least surprising option. Happy to close if you feel strongly it's out of scope

itzzdev09:
The framing on #321 was "do negative ordinals make sense," but that's not quite what this fixes. `ordinal()` already accepts negative ints today — it doesn't raise — it just returns grammatically impossible strings: `ordinal(-9)` gives `-9st`, `ordinal(-21)` gives `-21th`. So the choice isn't "support negatives or not," it's "keep emitting broken output, or fix it." If you'd genuinely rather not handle negatives, I'd argue raising or returning the value unchanged is cleaner than returning `-9st`. But given it already accepts them, correcting the suffix seemed like the least surprising option. Happy to close if you feel strongly it's out of scope.

hugovk:
Please don't let your LLM spam the same answer twice.

itzzdev09:
Sorry — that was a duplicate from a tooling mistake on my end, not intentional. Won't happen again.

